PR #8, opened 2017-07-19
Set environment for check commands

branden:
https://jira.mesosphere.com/browse/DCOS_OSS-1321

DC/OS PR: https://github.com/dcos/dcos/pull/1752

### DIFF
--- a/runner/check.go
+++ b/runner/check.go
@@ -26,7 +26,7 @@ type Check struct {
 }
 
 // Run executes the given check.
-func (c *Check) Run(ctx context.Context, role string) ([]byte, int, error) {
+func (c *Check) Run(ctx context.Context, role string, env []string) ([]byte, int, error) {
 	if !c.verifyRole(role) {
 		return nil, -1, errors.Errorf("check can be executed on a node with the following roles %s. Current role %s", c.Roles, role)
 	}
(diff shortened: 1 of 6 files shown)


runner/check.go
```diff
@@ -38,13 +38,15 @@ func (c *Check) Run(ctx context.Context, role string) ([]byte, int, error) {
...
-		timeout = time.Second*5
+		timeout = time.Second * 5
 	}
 
 	newCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
-	stdout, stderr, code, err := exec.Output(newCtx, c.Cmd...)
+	cmd := exec.CommandContext(newCtx, c.Cmd...)
+	cmd.Env = []string{"PATH=/bin:/usr/bin:/opt/mesosphere/bin"} // TODO(branden)
```
darkonie: maybe we should extend the original environment variables instead of overriding the whole thing?
i think this will remove the original and replace with only one variable PATH

Also, maybe we should make it a constant?

branden: This PR is a WIP, I should've put that in the title originally. I just wanted to make sure this'll build against dcos/dcos-go#30. Yeah we'll need to set more than `PATH`. The value of `PATH` itself will need to come from a config file, so that DC/OS can set it.

branden:
Ready for review!

PR #9, opened 2017-07-31
Mesos Authentication Support

matthewdfuller:


matthewdfuller:
We're still doing some testing with this, please don't merge yet.

api/diagnostics.go
```diff
@@ -499,6 +499,13 @@ func (j *DiagnosticsJob) getHTTPAddToZip(node Node, endpoints map[string]string,
 		}
 		request.Header.Add("Accept-Encoding", "gzip")
 
+		// Inject Mesos Auth where needed
+		if strings.Contains(fullURL, ':5050') || strings.Contains(fullURL, ':5051')  {
```
iderdik: These strings need to be wrapped in `"` (double quotes)

PR #11, opened 2017-08-25
Mesos Authentication Support

matthewdfuller:
Adds new flags for Mesos Auth User and Pass, allowing diagnostics to communicate with Mesos Master and Agents.


janisz:
Closing due to lack of activity

PR #12, opened 2017-09-19
Respect PATH env var when looking for check executables

branden:
This causes exec.FullOutput to respect the PATH env var when looking for the executable by setting the env when the runner process starts. The subprocess will then inherit the correct environment from the runner process.

### DIFF
--- a/runner/check.go
+++ b/runner/check.go
@@ -26,7 +26,7 @@ type Check struct {
 }
 
 // Run executes the given check.
-func (c *Check) Run(ctx context.Context, role string, env []string) ([]byte, int, error) {
+func (c *Check) Run(ctx context.Context, role string) ([]byte, int, error) {
 	if !c.verifyRole(role) {
 		return nil, -1, errors.Errorf("check can be executed on a node with the following roles %s. Current role %s", c.Roles, role)
 	}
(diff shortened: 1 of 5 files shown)


runner/check.go
```diff
@@ -44,9 +45,18 @@ func (c *Check) Run(ctx context.Context, role string, env []string) ([]byte, int
...
+		oldVal, set := os.LookupEnv(varName)
+		if set {
+			defer os.Setenv(varName, oldVal)
+		} else {
+			defer os.Unsetenv(varName)
+		}
+		os.Setenv(varName, newVal)
+	}
+
+	stdout, stderr, code, err := exec.FullOutput(exec.CommandContext(newCtx, c.Cmd...))
```
collinvandyck: I liked the version better where you were simply assigning `cmd.Env` to env.  That way you don't pollute the env for the entire process.  Would it be better to basically do something like this?

```
newEnv := []string
for _, e := range os.Environ() {
  newEnv = append(newEnv,e)
}
for k,v := range envOverride {
  newEnv = append(newEnv, k + "=" + v)
}
cmd := exec.CommandContext(newCtx, c.Cmd...)
cmd.Env = newEnv
....
```

branden: Me too, unfortunately I really do need to modify the parent process's env before starting the subprocess -- it's the parent process's `PATH` that gets used when searching for the child process's executable. If I went with the approach you suggest, the unit test I updated in this PR would fail because it can't find `combined.sh`.

I could have the runner config distinguish between parent and child process environments, but I couldn't come up with a good use case for it, so it seemed like unnecessary complexity.

branden:
retest this please

runner/check.go
branden: I changed this so that now the environment gets set when the runner process starts. This should eliminate any possible issues around concurrency.

branden:
I updated this PR to set the check environment in the runner's process when it starts up, but that means unit tests can't really cover this anymore. I'll ask for review again once there's a DC/OS PR that proves this works.

branden:
https://teamcity.mesosphere.io/viewLog.html?buildId=798426&buildTypeId=DcOs_Open_Test_IntegrationTest_InstallerCli&tab=buildResultsDiv proves that this works, so I'll go a head and merge.

PR #19, opened 2017-12-01
update error report if requested endpoint returned unexpected response code

mnaboka:
all endpoints must return 200, otherwise it should be an error

### DIFF
--- a/api/diagnostics.go
+++ b/api/diagnostics.go
@@ -507,6 +507,16 @@ func (j *DiagnosticsJob) getHTTPAddToZip(node Node, endpoints map[string]string,
 			j.JobProgressPercentage += percentPerURL
 			continue
 		}
+
+		if resp.StatusCode != http.StatusOK {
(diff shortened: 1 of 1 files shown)


api/diagnostics.go
```diff
@@ -507,6 +507,16 @@ func (j *DiagnosticsJob) getHTTPAddToZip(node Node, endpoints map[string]string,
 			j.JobProgressPercentage += percentPerURL
 			continue
 		}
+
+		if resp.StatusCode != http.StatusOK {
+			errMsg := fmt.Sprintf("unable to fetch %s. Return code %d", fullURL, resp.StatusCode)
+			logrus.Error(errMsg)
+			j.Errors = append(j.Errors, errMsg)
+			updateSummaryReport("", node, errMsg, summaryErrorsReport)
```
collinvandyck: Why do we use a blank string for the first argument to `updateSummaryReport` here?   The others seem to add the error message for this argument.

mnaboka: fixed

PR #20, opened 2017-12-21
make error message verbose

mnaboka:
if the dcos-diagnostics node responds with other then 200 status code,
we should make it clear in the logs.

### DIFF
--- a/api/pull.go
+++ b/api/pull.go
@@ -589,6 +589,14 @@ func pullHostStatus(host Node, respChan chan<- *httpResponse, dt *Dt, wg *sync.W
 	// use fake interface implementation for tests
 	timeout := time.Duration(dt.Cfg.FlagPullTimeoutSec) * time.Second
 	body, statusCode, err := dt.DtDCOSTools.Get(url, timeout)
+	if statusCode != http.StatusOK {
+		logrus.Errorf("Bad response code %d. URL %s", statusCode, url)
+		response.Status = statusCode
+		host.Health = 3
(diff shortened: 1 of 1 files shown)


api/pull.go
```diff
@@ -589,6 +589,14 @@ func pullHostStatus(host Node, respChan chan<- *httpResponse, dt *Dt, wg *sync.W
 	// use fake interface implementation for tests
 	timeout := time.Duration(dt.Cfg.FlagPullTimeoutSec) * time.Second
 	body, statusCode, err := dt.DtDCOSTools.Get(url, timeout)
+	if statusCode != http.StatusOK {
+		logrus.Errorf("Bad response code %d. URL %s", statusCode, url)
+		response.Status = statusCode
+		host.Health = 3
```
collinvandyck: is this a constant defined somewhere? should it be?

api/pull.go
```diff
@@ -589,6 +589,14 @@ func pullHostStatus(host Node, respChan chan<- *httpResponse, dt *Dt, wg *sync.W
...
+		response.Node = host
+		respChan <- &response
```
collinvandyck: Here you send a pointer to the response after you set response.Node=host.  Below you set response.Node = host after you send it.  does order matter here?

mnaboka: no, it does not matter

api/pull.go
mnaboka: it's not defined, there will be another PR to address it

PR #21, opened 2017-12-21
add errorsReport file to diagnostics bundle if not empty

mnaboka:
currently a diagnostics bundle will always include a file
`errorsSummaryReport.txt`, even if no errors reported. With this change
the file will be added only if not empty.

### DIFF
--- a/api/diagnostics.go
+++ b/api/diagnostics.go
@@ -219,24 +219,34 @@ func (j *DiagnosticsJob) runBackgroundJob(nodes []Node, cfg *config.Config, DCOS
 	// place a summaryErrorsReport.txt in a zip archive which should provide info what failed during the logs collection.
 	summaryErrorsReport := new(bytes.Buffer)
 	defer func() {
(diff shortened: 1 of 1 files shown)


mnaboka:
@collinvandyck fixed

api/diagnostics.go
```diff
@@ -219,17 +219,21 @@ func (j *DiagnosticsJob) runBackgroundJob(nodes []Node, cfg *config.Config, DCOS
...
+		if summaryErrorsReport.Len() > 0 {
+			zipFile, err := zipWriter.Create("summaryErrorsReport.txt")
+			if err != nil {
+				j.Status = "Could not append a summaryErrorsReport.txt to a zip file"
+				logrus.Errorf("%s: %s", j.Status, err)
+				j.Errors = append(j.Errors, err.Error())
+				return
+			}
+
+			io.Copy(zipFile, summaryErrorsReport)
```
collinvandyck: we should probably not ignore the error from `io.Copy` either.

mnaboka: good catch, fixed

PR #25, opened 2018-02-23
Update Makefile to build on Mac and Linux

collinvandyck:
JIRA: https://jira.mesosphere.com/browse/DCOS-21365

In short, the goal is to make our go projects buildable on both the Mac and Linux platforms.


collinvandyck:
I also had to change the Jenkins build to just run `make test` instead of trying to run it in Docker (since `make test` ends up running it in Docker to begin with).  I will end up modifying any sort of build that builds master in the same way.

PR #31, opened 2018-03-08
runner: treat command timeout as a failed check

mnaboka:
[DCOS_OSS-2247](https://jira.mesosphere.com/browse/DCOS_OSS-2247)

there are several levels of errors when runner executes checks. The error
is considered to be an exception, where runner is unable to run the checks for
some reason. For example, when the check is not present in the path.
However if the command execution takes more time then defined in `timeout`
check property, it will be killed and also considered as an error. But in fact,
it should be a failed check. This PR fixes it.

### DIFF
--- a/runner/check.go
+++ b/runner/check.go
@@ -2,6 +2,8 @@ package runner
 
 import (
 	"context"
+	"fmt"
+	goexec "os/exec"
 	"runtime"
 	"time"
 
@@ -9,6 +11,9 @@ import (
 	"github.com/dcos/dcos-go/exec"
 	"github.com/pkg/errors"
 )
+
+// signalKilled is an error message returned by dcos-go/exec if the check exceeds timeout
+const signalKilled = "signal: killed"
 
 // Check is a basic structure that describes DC/OS check.
 type Check struct {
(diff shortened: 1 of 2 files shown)


runner/check.go
```diff
@@ -46,6 +51,13 @@ func (c *Check) Run(ctx context.Context, role string) ([]byte, int, error) {
 
 	stdout, stderr, code, err := exec.FullOutput(exec.CommandContext(newCtx, c.Cmd...))
 	if err != nil {
```
branden: Could we assume at this point that any error is a check failure, rather than a runner error? If we check our prerequisites that might result in runner errors before we attempt to run the command, then assume that if the command fails it's a check failure rather than an error, we won't need to do potentially brittle things like inspect output for a specific string.

mnaboka: well, we could do that, however it would mean we are breaking APIs defined in design doc regarding the *real* errors output.

It just unfortunate that the golang `os./exec` module sends an error on timeout.

bernadinm: To @branden point, I've managed to create an ip-detect script that will fail in the condition of a successful return of the contents of the ip-detect script when run, even when the exit code is returned as 22 rather than zero.

PR #33, opened 2018-04-12
Run checks in parallel

amitaekbote:


runner/runner.go
```diff
@@ -231,60 +264,77 @@ func (r *Runner) run(ctx context.Context, checkMap map[string]*Check, list bool,
...
+	responseChannel := make(chan *Response, len(currentCheckList))
+	wg := &sync.WaitGroup{}
+	
 	// main loop to get the checks info.
 	for _, name := range currentCheckList {
-		resp := &Response{
-			name: name,
-		}
+		wg.Add(1)
+		xname := name
```
gpaul: If you do this you don't need to bind the variable as an argument to the closure on the next line - and conversely, if you pass `name` to the closure below on line 335, you don't need to do this local showing trick

runner/runner.go
```diff
@@ -231,60 +264,77 @@ func (r *Runner) run(ctx context.Context, checkMap map[string]*Check, list bool,
...
+		go func(xname string, responseChannel chan *Response) {
+			defer wg.Done()
+			currentCheck, ok := checkMap[xname]
+			if !ok {
+				combinedResponse.errs["check not found"] = &Response{
```
gpaul: Pretty sure this is a race condition. You are assigning to `errs` from multiple goroutines. The coolest way to fix this IMO is to add a method on combinedResponse called `SetErr()` or similar and have that method take a mutex lock when it enters and defers the unlock.

runner/runner.go
```diff
@@ -231,60 +264,77 @@ func (r *Runner) run(ctx context.Context, checkMap map[string]*Check, list bool,
...
+			if err != nil {
+				combinedResponse.errs[err.Error()] = resp
+			}
 
-		combinedResponse.status = max(combinedResponse.status, code)
+			combinedResponse.status = max(combinedResponse.status, code)
+			
+		}(xname, responseChannel)
 	}
+	wg.Wait()
```
gpaul: What do you think about the following alternative?

```
type responseError struct {
  err error
  checkName string
}

errs := make(chan responseError)
responses := make(chan error)

for _, name := range currentCheckList {
  go func(name string) {
    defer wg.Done()
    currentCheck, ok := checkMap[name]
    if !ok {
      errs <- responseError{errors.New("check not found"), name}
      return
    }
    ...
    // find runner for the given role only.
    if !currentCheck.verifyRole(r.role) {
      // send an empty response for this check, as it is not relevant for this role.
      responses <- nil
      return
    }
    ...
    resp := &Response{ ... }
    ...
    responses <- resp
    return
  }(name)
}

// Expect one response or error for every check
for range currentCheckList {
  select {
  case err := <-errs:
    combinedResponse.errs[err.Error()] = err.name
  case resp := <-responses:
    if resp == nil {
      // skip filtered checks that are not appropriate for this role.
      continue
    }
    combinedResponse.checks[resp.name] = resp
  }
}

combinedResponse.status = ...
return combinedResponse, nil
```

klueska: This is how I would do it

runner/runner.go
```diff
@@ -231,59 +240,92 @@ func (r *Runner) run(ctx context.Context, checkMap map[string]*Check, list bool,
...
+				return
+			}
 
-		var (
-			combinedOutput []byte
-			code           int
-			err            error
-			checkDuration  string
-		)
+			if _, ok := combinedResponse.checks[name]; ok {
```
gpaul: Maybe we should move this check to the for loop by deduplicating `currentCheckList` before we start looping?

runner/runner.go
```diff
@@ -55,6 +55,14 @@ type responseList struct {
 	Timeout     string   `json:"timeout"`
 }
 
+type responseError struct {
+	err           string
+	checkName     string
+	checkNotFound bool
+	response      *Response
+
```
mnaboka: redundant space

gpaul:
@amitaekbote is this ready for another pass?

amitaekbote:
yes it is

runner/runner.go
```diff
@@ -231,59 +240,85 @@ func (r *Runner) run(ctx context.Context, checkMap map[string]*Check, list bool,
 		}
 	}
 
+	errs := make(chan responseError, len(currentCheckList))
+	responses := make(chan *Response, len(currentCheckList))
```
gpaul: As `responseError` contains a `*Response`, could we get rid of the `responses` channel and rename `errs` to `results` or something similar? Then we have only one channel. (I believe you suggested this a while ago.)

runner/runner.go
gpaul: Checking `combinedResponse.checks` here is a data race: the on lines 303+ modify this map. I think we should just range through `currentCheckList` on line 246 (before entering the for loop) and remove duplicates.

runner/runner.go
```diff
@@ -231,59 +240,85 @@ func (r *Runner) run(ctx context.Context, checkMap map[string]*Check, list bool,
...
+			// collect errors
+			if err != nil {
+				errs <- responseError{err.Error(), name, false, resp}
+			}
+		}(name)
+	}
+
+	for range currentCheckList {
+		select {
+		case err := <- errs:
```
gpaul: if we rename `errs` to `results` and remove `responses` in favour of using `resultError.response` then we can merge the two `case` statements here.

branden:
I ported this PR to the new check runner component: https://github.com/dcos/dcos-check-runner/pull/3. @amitaekbote You can go ahead and close this out.

PR #36, opened 2018-05-30
convert vendor folder to godep and update all packages

mnaboka:
- remove vendor folder from govendor
- use godep
- bump all packages to the latest version
- replace `github.com/Sirupsen/logrus` with `github.com/sirupsen/logrus`


janisz:
Closing in favour of #43

PR #42, opened 2018-07-25
Add health mapping to readme

DanielMSchmidt:
Hi there,

I thought I should add the information I got by asking to the place I anticipated it to be so that other clueless people like me may find the same info easier.

Feel free to reject if it's not helpful at all :)


philipnrmn:
Thanks for the PR @DanielMSchmidt

DanielMSchmidt:
You are very welcome, thank you for helping me earlier :)

PR #47, opened 2018-08-23
Minor code improvements

janisz:
Currently primitive types are used for any domain objecst
in the system. This change intorduce Health type which is
aliased `int`. See: [PrimitiveObsession](http://wiki.c2.com/?PrimitiveObsession)

There is no need ot expose base route outside of the package.

This structure is implemented in different place.
There is no need to keep it.

### DIFF
--- a/api/handlers_test.go
+++ b/api/handlers_test.go
@@ -127,7 +127,7 @@ func (st *fakeDCOSTools) Get(url string, timeout time.Duration) (body []byte, st
 	}
 	var response string
 	// master
-	if url == fmt.Sprintf("http://127.0.0.1:1050%s", BaseRoute) {
+	if url == fmt.Sprintf("http://127.0.0.1:1050%s", baseRoute) {
 		response = `
 			{
 			  "units": [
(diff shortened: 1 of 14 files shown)


api/structures.go
```diff
@@ -16,11 +16,19 @@ type MonitoringResponse struct {
 	UpdatedTime time.Time
 }
 
+// Health is a type to indicates health of the unit.
+type Health int
+
+const (
+	unhealthy = 0
```
klueska: Is there a reason to make these all lower case rather than all CamelCase, like for most other constants? Is it to keep it hidden in the api package?

klueska: It just feels a bit odd to have package level constants called simply `healthy` and `unhealthy`

janisz: Youre right. Exported type constants should be also exported.

api/handlers_test.go
```diff
@@ -19,17 +19,16 @@ import (
 	"github.com/stretchr/testify/suite"
 )
 
-var DiagnosticsBundleDir = "/tmp/snapshot-test"
-var testCfg *config.Config
 
-func init() {
-	if runtime.GOOS == "windows" {
```
ArmandGrillet: Just to be sure: using `ioutil.TempDir` you remove the need of OS dependent code, right?

janisz: Correct :heavy_check_mark:

PR #48, opened 2018-08-28
Do not sleep in tests

janisz:
Change `time.Sleep` to busy waiting for bundle to complete.

### DIFF
--- a/.gometalinter.json
+++ b/.gometalinter.json
@@ -3,7 +3,7 @@
   "Concurrency": 4,
   "LineLength": 150,
   "Cyclo": 20,
-  "Deadline": "300s",
+  "Deadline": "600s",
   "DisableAll": true,
   "Enable": [
     "deadcode",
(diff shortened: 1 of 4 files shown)


janisz:
Tests on Windows fails due to race. It looks like this whole code is buggy. At first I thought only `Running` and `Errors` may be problematic but other filed are also updated (`Status`). I need to figure out how to fix it. Probably the best option is to extract state from `DiagnosticJob` and allow it's modification via synchronized methods or channels.
 @klueska @philipnrmn  What do you think?

janisz:
I run this tests 100 times on Linux and could not reproduce the problem

```
TRIES=100
COMMAND="GOCACHE=off go test ./api -race "
n=0
until [ $n -ge $TRIES ]
do
  eval ${COMMAND} || break
  n=$[$n+1]
done
```

klueska:
Trigger windows tests

PR #52, opened 2018-09-10
Fix broken logic in tests

janisz:
PRevious commit f091937393eea045d90c92be59abacdfb436e252 introduced
checking for bundle clomplete. Unfortunatelly it was not a real fix
since all requests were mocked. This PR changes this logic to use
filesystem instead of usign API. In the future we should add integration
test in for this project where we could test whole flow and interact
with API instead of filestystem.


api/diagnostics_test.go
```diff
@@ -169,10 +168,12 @@ func TestGetAllStatus(t *testing.T) {
 
 func TestIsSnapshotAvailable(t *testing.T) {
 	tools := &fakeDCOSTools{}
+	cfg := testCfg()
+	defer os.Remove(cfg.FlagDiagnosticsBundleDir)
```
bamarni: Is it possible that this directory contains files? If so `RemoveAll` would be more robust.

api/diagnostics_test.go
```diff
@@ -426,44 +423,24 @@ func TestRunSnapshot(t *testing.T) {
...
+func waitForBundle(t *testing.T, bundlePath string) bool {
 	timeout := time.After(2 * time.Second)
 	for {
 		select {
 		case <-timeout:
 			t.Log("Timeout!")
 			return false
 		default:
-			response, code, err := MakeHTTPRequest(t, router, "http://127.0.0.1:1050/system/health/v1/report/diagnostics/status", "GET", nil)
+			stat, err := os.Stat(bundlePath)
```
bamarni: Even though it is a lighter operation than what it used to be, in these situations I'd add a small sleep after this instead of looping as fast as possible while polling for the file.

api/handlers_test.go
```diff
@@ -19,18 +19,18 @@ import (
...
 	return &config.Config{
-		FlagRole:                 "master",
-		FlagDiagnosticsBundleDir: diagnosticsBundleDir,
-		FlagPort:                 1050,
-		FlagMasterPort:           1050,
+		FlagRole:                         "master",
+		FlagDiagnosticsBundleDir:         diagnosticsBundleDir,
+		FlagPort:                         1050,
+		FlagMasterPort:                   1050,
+		FlagDiagnosticsJobTimeoutMinutes: 1,
```
bamarni: not directly related to the PR, but it'd be good to have it called `FlagDiagnosticsJobTimeout` and it'd take a `time.Duration`

janisz: I agree but then it could change its semantic in `config.json` Ideally I'd like to create custom structure and handle it's serialization.

janisz:
@bamarni Fixed

PR #55, opened 2018-09-14
Use filepath.Base instead of path.Base

janisz:
`path.Base` has a problem when the path contains drive. To fix this problem I switched to `filepath.Base` that return dir and filename.

See golang/go#7698


janisz:
@gpaul I copied description from commit message https://github.com/dcos/dcos-diagnostics/commit/31c614572f6d2d8ad3667b4373b9397c06f3cfdc

gpaul:
What is the problem exactly?

janisz:
Build not pass on Windows because temp directory where we store bundles has drive in it so `filetpah.Base` returns wole path not only file name.

gpaul:
See https://github.com/golang/go/issues/7698

The issue is that you're using `path.Base` instead of `filepath.Base`.

PR #56, opened 2018-09-20
Fail when no external log providers config files

janisz:
Current implementation silently goes over errors with reading `endpoint-config`. This commit changes this behaviour to fail when external config could not be loaded. No matter if the file is empty, does not exist or is invalid JSON.

Fixes: [DCOS_OSS-4130](https://jira.mesosphere.com/browse/DCOS_OSS-4130)

### DIFF
--- a/api/diagnostics.go
+++ b/api/diagnostics.go
@@ -803,16 +803,26 @@ type CommandProvider struct {
 	indexedCommand string
 }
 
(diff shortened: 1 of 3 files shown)


janisz:
@gpaul I like this idea. I'll do it in the next PR. Ideally, the whole configuration should be read and validated at startup.

PR #60, opened 2018-09-26
Add tests for DiagnosticsJob

janisz:
Currenlty, the code has no tests. In this PR I added a bunch of tests for `DiagnosticsJob` and cleanup code a little bit.

### DIFF
--- a/api/diagnostics.go
+++ b/api/diagnostics.go
@@ -310,7 +310,8 @@ func (j *DiagnosticsJob) runBackgroundJob(nodes []Node) {
 			continue
 		}
 		// add http endpoints
-		err = j.getHTTPAddToZip(node, endpoints, zipWriter, summaryErrorsReport, summaryReport, percentPerNode)
+		err = j.getHTTPAddToZip(node, endpoints, zipWriter, summaryErrorsReport,
+			summaryReport, percentPerNode)
 		if err != nil {
 			j.Errors = append(j.Errors, err.Error())
 
@@ -341,7 +342,7 @@ func (j *DiagnosticsJob) runBackgroundJob(nodes []Node) {
 // delete a bundle
 func (j *DiagnosticsJob) delete(bundleName string) (response diagnosticsReportResponse, err error) {
 	if !strings.HasPrefix(bundleName, "bundle-") || !strings.HasSuffix(bundleName, ".zip") {
-		return prepareResponseWithErr(http.StatusServiceUnavailable, errors.New("format allowed  bundle-*.zip"))
+		return prepareResponseWithErr(http.StatusBadRequest, errors.New("format allowed  bundle-*.zip"))
 	}
 
 	j.Lock()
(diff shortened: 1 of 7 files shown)


api/diagnostics.go
```diff
@@ -484,14 +484,13 @@ func (d diagnosticsJobCanceledError) Error() string {
 }
 
 // fetch an HTTP endpoint and append the output to a zip file.
-func (j *DiagnosticsJob) getHTTPAddToZip(node Node, endpoints map[string]string, folder string, zipWriter *zip.Writer,
```
philipnrmn: How did this ever build? I would expect errors on the unused parameter...

api/diagnostics.go
```diff
@@ -518,32 +517,11 @@ func (j *DiagnosticsJob) getHTTPAddToZip(node Node, endpoints map[string]string,
...
-
-		if resp.StatusCode != http.StatusOK {
-			errMsg := fmt.Sprintf("unable to fetch %s. Return code %d", fullURL, resp.StatusCode)
-			logrus.Error(errMsg)
-			j.Errors = append(j.Errors, errMsg)
-			updateSummaryReport("wrong status code returned", node, errMsg, summaryErrorsReport)
+			logrus.Errorf("Could not read force-tls flag: %s", err)
+			updateSummaryReport(err.Error(), node, err.Error(), summaryErrorsReport)
 			j.JobProgressPercentage += percentPerURL
 			continue
```
philipnrmn: We don't close the body in this case; isn't this going to lead to hanging connections?

gpaul: If you close the response body here 
https://github.com/dcos/dcos-diagnostics/pull/60/files#diff-0ee9de7b9d9927cce9fb53251662d72aR564 (so if the request succeeded but the response is unexpected) then the `get()` function maintains a contract that says if it returns an error, the response body does not need to be closed. 

That's a convenient pattern to follow.

api/diagnostics.go
janisz: Good point. The check was disabled. I fixed it here https://github.com/dcos/dcos-diagnostics/pull/62

api/diagnostics.go

janisz:
@philipnrmn Fixed

api/diagnostics.go
```diff
@@ -485,12 +485,11 @@ func (d diagnosticsJobCanceledError) Error() string {
 // fetch an HTTP endpoint and append the output to a zip file.
 func (j *DiagnosticsJob) getHTTPAddToZip(node Node, endpoints map[string]string, zipWriter *zip.Writer,
 	summaryErrorsReport, summaryReport *bytes.Buffer, percentPerNode float32) error {
-	if len(endpoints) == 0 || percentPerNode == 0 {
```
gracedo: just wondering why these checks can be removed? (we don't need to worry about a divide by 0 panic?)

api/diagnostics.go
```diff
@@ -517,32 +516,11 @@ func (j *DiagnosticsJob) getHTTPAddToZip(node Node, endpoints map[string]string,
...
-			j.JobProgressPercentage += percentPerURL
-			continue
-		}
-
-		if resp.StatusCode != http.StatusOK {
-			errMsg := fmt.Sprintf("unable to fetch %s. Return code %d", fullURL, resp.StatusCode)
-			logrus.Error(errMsg)
-			j.Errors = append(j.Errors, errMsg)
-			updateSummaryReport("wrong status code returned", node, errMsg, summaryErrorsReport)
+			logrus.Errorf("Could not read force-tls flag: %s", err)
```
gracedo: this blanket error message doesn't look quite right

api/diagnostics.go
janisz: It's covered here https://github.com/dcos/dcos-diagnostics/pull/60/files#diff-0ee9de7b9d9927cce9fb53251662d72aR303 (Github fold this part).

api/diagnostics.go
janisz: Good point

api/diagnostics.go
janisz: I mocked `/logs` endpoint in one tests to validate this case.

janisz:
@philipnrmn @gracedo Can you look one more time?

PR #62, opened 2018-09-27
Remove unused parameters and fix HTTP codes

janisz:
For any error, we returned HTTP 503 which was improper when
user send us Bad Request.

### DIFF
--- a/.gometalinter.json
+++ b/.gometalinter.json
@@ -22,6 +22,7 @@
     "goconst",
     "lll",
     "misspell",
+    "unparam",
     "unused",
     "vet"
   ],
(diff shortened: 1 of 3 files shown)


api/handlers.go
```diff
@@ -265,7 +265,7 @@ func diagnosticsJobStatusAllHandler(w http.ResponseWriter, r *http.Request) {
 func cancelBundleReportHandler(w http.ResponseWriter, r *http.Request) {
 	dt, ok := getDtFromContext(r.Context())
 	if !ok {
-		httpError(w, "Unable to get a required context from a request", http.StatusInternalServerError)
+		httpError(w, "Unable to get a required context from a request", http.StatusBadRequest)
```
philipnrmn: Can we make these errors more descriptive please? I won't block the PR on it but it'd be really nice :)

.gometalinter.json
```diff
@@ -22,6 +22,7 @@
     "goconst",
     "lll",
     "misspell",
+    "unparam",
```
philipnrmn: 🙌

api/handlers.go
janisz: Sure I have one PR to fix config and fixing this error codes will be next. I'll also add tests for it. My goal is to get health coverage (about 70%) and then start changing the behaviour.

PR #65, opened 2018-10-02
Create dcos package

janisz:
This change does NOT change the behaviour only moves code to the following packages:

* `utils` – common helpers (functions with small logic) that are shared between packages.
* `unit` – everything related with systemd interaction
* `dcos` – everything that touches other DCOS elements (e.g., Nodes)


api/propertes_windows.go
```diff
@@ -0,0 +1,18 @@
+package api
```
gracedo: typo in file name `propertes` -> `properties`

cmd/daemon.go
```diff
@@ -39,7 +40,7 @@ const (
 // override the defaultStateURL to use https scheme
 var defaultStateURL = url.URL{
 	Scheme: "https",
-	Host:   net.JoinHostPort(dcos.DNSRecordLeader, strconv.Itoa(dcos.PortMesosMaster)),
+	Host:   net.JoinHostPort(dcosGo.DNSRecordLeader, strconv.Itoa(dcosGo.PortMesosMaster)),
```
gracedo: it looks like some constants are pulled out of `github.com/dcos/dcos-go/dcos` (e.g. `RoleMaster`) into the `dcos` package in this project - should `DNSRecordLeader` and `PortMesosMaster` be as well to avoid importing both of these `dcos` pkgs here?

api/propertes_windows.go
branden: +1, please fix the typo in the filename here.

cmd/daemon.go
janisz: Good point. I moved the whole definition of `defaultStateUrl` to dcos package.

janisz:
@gracedo @branden Thank you! I fixed all issues :)

PR #67, opened 2018-10-18
Extract common code to a function in pull.go

janisz:
Extracted common code to a function and use structured logs.
Log response body when status code is not 200.
Introduce `Unknown` value for `dcos.Health`

### DIFF
--- a/api/pull.go
+++ b/api/pull.go
@@ -106,13 +106,18 @@ func updateHealthStatus(responses <-chan *httpResponse, dt *Dt) {
 func pullHostStatus(host dcos.Node, respChan chan<- *httpResponse, dt *Dt, wg *sync.WaitGroup) {
 	defer wg.Done()
 	var response httpResponse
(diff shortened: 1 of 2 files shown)


api/pull.go
```diff
@@ -134,30 +136,21 @@ func pullHostStatus(host dcos.Node, respChan chan<- *httpResponse, dt *Dt, wg *s
...
+		markNodeHealthAsUnknown(statusCode)
 		return
 	}
 	if err != nil {
-		logrus.Errorf("Could not HTTP GET %s: %s", url, err)
-		response.Status = statusCode
-		host.Health = 3 // 3 stands for unknown
-		respChan <- &response
-		response.Node = host
+		logrus.WithError(err).WithField("URL", url).Errorf("Could not HTTP GET")
```
gracedo: this can just be `.Error`

janisz: Thanks

janisz:
@gracedo It's fixed

PR #70, opened 2018-10-22
Remove dependency on `Dt` and `MR` in diagnostics

janisz:
Diagnostic Job is dependent on Dt and MR while it has all the required data. This commit untangles the code.

This commit is also the first step to start using mock lib instead of writing it by hand.


api/mock_tools_test.go
```diff
@@ -0,0 +1,81 @@
+package api
```
gracedo: do you use an autogenerator such as https://github.com/vektra/mockery to make these mocks? I wonder if it's worth pulling into the Makefile so that all mocks in the repo can be regenerated in one make target (e.g. if any of the interfaces change it would be pretty easy to use the make target to regenerate the mocks)

janisz: Thanks, I've never used mockery. I used "Implement interface" IntelliJ feature. Once we get more mocked objects I'll use mockery to generate code.

PR #71, opened 2018-10-25
`UseTLSScheme` returns error when URL is empty

janisz:
This is a fix for problems with building dcos-diagnostics on Windows.

See: https://jenkins.mesosphere.com/service/jenkins/job/public-dcos-diagnostics-windows/127/console
Refs: #68


janisz:
@oprinmarius  can you take a look?

PR #73, opened 2018-10-26
Create pull structure and remove dependecy on `Dt`

janisz:


api/pull.go
```diff
@@ -13,33 +13,48 @@ import (
 	"github.com/sirupsen/logrus"
 )
```
philipnrmn:
```suggestion
// pull represents process that fetches a complete health status response
```

PR #75, opened 2018-10-31
Use 404 when unit or node could not be found

janisz:


api/handlers.go
```diff
@@ -52,9 +52,14 @@ func (h *handler) getAllUnitsHandler(w http.ResponseWriter, _ *http.Request) {
 // /api/v1/system/health/units/:unit_id:
 func (h *handler) getUnitByIDHandler(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
-	unitResponse, err := h.monitoringResponse.GetUnit(vars["unitid"])
+	unitId := vars["unitid"]
```
gracedo: just wondering why this was pulled out into a var (doesn't look to be used elsewhere)

api/handlers.go
```diff
@@ -78,10 +88,15 @@ func (h *handler) getNodesByUnitIDHandler(w http.ResponseWriter, r *http.Request
 // /api/v1/system/health/units/:unit_id:/nodes/:node_id:
 func (h *handler) getNodeByUnitIDNodeIDHandler(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
-	nodePerUnit, err := h.monitoringResponse.GetSpecificNodeForUnit(vars["unitid"], vars["nodeid"])
+	unitId := vars["unitid"]
+	nodePerUnit, err := h.monitoringResponse.GetSpecificNodeForUnit(unitId, vars["nodeid"])
```
gracedo: it's a little inconsistent that `unitId` is pulled out into a var but not `vars["nodeid"]` - I think maybe it was ok to leave it as originally written?

janisz: Fixed

api/handlers.go
janisz: Accidentally committed my WIP version. Thanks for pointing out.

PR #76, opened 2018-11-21
Do NOT make remote call for local status

janisz:
Currently to get status we query all nodes including local one.
This change try to get local status using local object.

### DIFF
--- a/api/diagnostics_test.go
+++ b/api/diagnostics_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/dcos/dcos-diagnostics/dcos"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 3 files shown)


api/diagnostics_test.go
```diff
@@ -392,6 +401,64 @@ func TestGetAllStatus(t *testing.T) {
...
+		JobDuration:                              "2s",
+		DiagnosticBundlesBaseDir:                 "/home/core/1",
+		DiagnosticsJobTimeoutMin:                 720,
+		DiskUsedPercent:                          28.0,
+		DiagnosticsUnitsLogsSinceHours:           "24",
+		DiagnosticsJobGetSingleURLTimeoutMinutes: 5,
+		CommandExecTimeoutSec:                    10,
+	})
+
+	tools.AssertExpectations(t)
```
mhrabovcin: What expectations are set on this mock?

janisz: that only mocked function was called with expected arguments

api/diagnostics_test.go
```diff
@@ -392,6 +401,64 @@ func TestGetAllStatus(t *testing.T) {
...
+			}`
+
+	tools.On("Get",
+		mock.MatchedBy(func(url string) bool {
+			return url == fmt.Sprintf("http://127.0.0.2:1050%s/report/diagnostics/status", baseRoute)
+		}),
+		mock.MatchedBy(func(t time.Duration) bool { return t == 3*time.Second }),
+	).Return([]byte(mockedResponse), http.StatusOK, nil)
+	tools.On("DetectIP").Return("127.0.0.1", nil)
+	tools.On("GetMasterNodes").Return([]dcos.Node{{Leader: true, IP: "127.0.0.2", Role: "master"}}, nil)
```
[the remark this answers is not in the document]
janisz: That's right, thanks!

PR #79, opened 2019-01-08
Remove indexes from command names.

janisz:
There is no reason to keep this indexes and nobody remember why it was introduced. Probably to prevent name conflict when running the same command twice (but why we would do this?).

This change jsut remove index. All logic and structures remains unchanged.

**Fixes:** [DCOS_OSS-4155](https://jira.mesosphere.com/browse/DCOS_OSS-4155)

**See:**

* https://github.com/dcos/dcos/pull/3446#discussion_r239251129
* https://github.com/dcos/dcos/pull/3212#issuecomment-418964030

### DIFF
--- a/api/diagnostics.go
+++ b/api/diagnostics.go
@@ -804,12 +804,12 @@ func (j *DiagnosticsJob) getLogsEndpoints() (endpoints map[string]string, err er
 	}
 
 	// command endpoints
-	for indexedCommand, c := range j.logProviders.LocalCommands {
+	for cmdKey, c := range j.logProviders.LocalCommands {
 		if !roleMatched(currentRole, c.Role) {
 			continue
 		}
-		if indexedCommand != "" {
-			endpoints[indexedCommand] = fmt.Sprintf(":%d%s/logs/cmds/%s", port, baseRoute, indexedCommand)
+		if cmdKey != "" {
+			endpoints[cmdKey] = fmt.Sprintf(":%d%s/logs/cmds/%s", port, baseRoute, cmdKey)
 		}
 	}
 	return endpoints, nil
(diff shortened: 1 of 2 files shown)


adamtheturtle:
> Probably to prevent name conflict when running the same command twice (but why we would do this?).

Which command will now not work if run twice?
Can we have a test which shows that?

> Question is: can we do this? Or will this break some existing behaviour that customers rely on? Or other tools rely on?

I have the same concern.
Unfortunately the best way to know this is to have a test which demonstrates the reasoning behind having the existing behaviour.
My thinking is that we cannot forever protect code that does not have this documentation / testing else we will be paralysed.
Personally, I am OK with the risk here, if @janisz commits to staying on top of this and fixing any broken behaviour as we find out about it, and adding the required tests.

First, I will look at the git blame for the old lines and see if I can find any reasoning for the indexes.

janisz:
> First, I will look at the git blame for the old lines and see if I can find any reasoning for the indexes.

Done that. Unfortunately, the code was written by @mnaboka with no PRs, review so there is no information why it was done like this. I asked him on [DC/OS slack](https://screenshots.firefox.com/TBOu9XfFzjOJvIjD/dcos-community.slack.com) and it looks like it was never needed.

> Which command will now not work if run twice?

If you have the same command twice in configuration, the only one output will appear in the bundle. There is a small chance that two commands will have the same file name eg `ifconfig_eth0` and `ifconfig eth0` but it's very unlikely

> can we do this? Or will this break some existing behaviour that customers rely on? Or other tools rely on?

Yes, dcos-diagnostics is a tool to debug customer issues. Now every time a command is added/changed, filename changes so I think nobody relies on that.

PR #81, opened 2019-01-09
Updating expected results to include carriage return when run on windows

pcrews:
Attempting to update tests to pass after rebuilding our windows 10 agents.
Test is current failing due to expected results not including the '\r' character.
Am open to alternatives if someone can suggest a means to configure win10 to not produce '\r' characters in output.

### DIFF
--- a/api/diagnostics_test.go
+++ b/api/diagnostics_test.go
@@ -140,7 +140,7 @@ func TestDispatchLogsForCommand(t *testing.T) {
 
 	data, err := ioutil.ReadAll(r)
 	require.NoError(t, err)
-        var expected_result string
+        var expectedResult string
         expectedResult = "OK\n"
         if runtime.GOOS == "windows" {
            expectedResult = "OK\r\n"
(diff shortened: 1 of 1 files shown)


api/diagnostics_test.go
```diff
@@ -140,7 +140,7 @@ func TestDispatchLogsForCommand(t *testing.T) {
 
 	data, err := ioutil.ReadAll(r)
 	require.NoError(t, err)
-	assert.Equal(t, "OK\n", string(data))
+	assert.Equal(t, "OK\r\n", string(data))
```
janisz: Now it's failing here https://ci.appveyor.com/project/janisz/dcos-diagnostics/builds/21482670

mrnugget:
Hey! Do you think there's either
a) a platform-independent way to write this assertion or
b) a way to change the assertion at run-time so that it checks for the correct line break depending on which platform the tests are run?

Because now the tests are passing on Windows, but they're failing on Linux :)

api/diagnostics_test.go
pcrews: I have updated with an 'if' statement that should help clarify shifts in expected results and work on all platforms.

janisz:
@pcrews it's not working on AppVeyor now

PR #82, opened 2019-01-15
Fix compilation on Darwin by adding noop stubs of interfaces

mrnugget:
This adds the necessary noop functions to allow compilation on Darwin, which makes it possible to run tests/linters/etc.

When possible I tried to copy the behaviour of the Windows-implementations and where not I opted to return an error to signal that it's an unsupported feature.

I'm still unsure about the comments, opinions welcome 😄

### DIFF
--- a/api/diagnostics_test.go
+++ b/api/diagnostics_test.go
@@ -164,7 +164,7 @@ func TestDispatchLogsForFiles(t *testing.T) {
 }
 
 func TestDispatchLogsForUnit(t *testing.T) {
-	if runtime.GOOS == "windows" {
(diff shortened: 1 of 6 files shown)


janisz:
Can we have a CI for darwin? I'm not using the mac and I have no place to check if my changes break anything there.

mrnugget:
> Can we have a CI for darwin? I'm not using the mac and I have no place to check if my changes break anything there.

I think that's more trouble than it's worth. And I don't think you'll break something by accident that's not easily fixed by another developer that's using Darwin/macOS. But I think it would make sense to always add the platform-specific noop stubs for Darwin too, as a general rule.

PR #83, opened 2019-01-16
Replace blocking operation

janisz:
In the current implementation, we block whole `DcosDiagnosticsJob` when the bundle was generated. This PR changes this and lock only mutexes that are created per field: `Status`, `Errors`, `JobProgressPercentage`. Ater this change we can query for the bundle status even if we are waiting for some long operation.


janisz:

api/diagnostics.go
```diff
@@ -514,42 +523,53 @@ func (j *DiagnosticsJob) getHTTPAddToZip(node dcos.Node, endpoints map[string]st
...
-		resp.Body.Close()
-		updateSummaryReport("STOP "+j.Status, node, "", summaryReport)
-		j.JobProgressPercentage += percentPerURL
+func (j *DiagnosticsJob) getDataToZip(node dcos.Node, httpEndpoint string, client *http.Client, fileName string, zipWriter *zip.Writer) error {
+	fullURL, err := util.UseTLSScheme("http://"+node.IP+httpEndpoint, j.Cfg.FlagForceTLS)
+	if err != nil {
+		e := fmt.Errorf("could not read force-tls flag: %s", err)
+		return e
+	}
+	resp, err := get(client, fullURL)
```
mrnugget: Since you now set `client` on `*DiagnosticJob` you can reuse it here and don't need to pass it around

api/diagnostics.go
```diff
@@ -443,7 +464,7 @@ func (j *DiagnosticsJob) getStatusAll() (map[string]bundleReportStatus, error) {
 }
 
 // get a status report for a localhost
-func (j *DiagnosticsJob) getStatus() bundleReportStatus {
+func (j *DiagnosticsJob) getBundelReportStatus() bundleReportStatus {
```
mrnugget: Typo: `Bundel` -> `Bundle`

PR #86, opened 2019-02-06
Raise HTTP timeout on call to Mesos to 30 seconds from 1 in line with Tim's advice

br-lewis:


dcos/finder.go
```diff
@@ -225,7 +225,7 @@ func (f *findNodesInDNS) getMesosAgents() (nodes []Node, err error) {
 		return nodes, err
 	}
 
-	timeout := time.Second
+	timeout := 30 * time.Second
```
mrnugget: Can we pull this in a constant, now that we're touching it anyway?

PR #87, opened 2019-02-12
Log response body from remote instance

janisz:


api/diagnostics.go
```diff
@@ -519,7 +519,11 @@ func (j *DiagnosticsJob) getHTTPAddToZip(node Node, endpoints map[string]string,
 		}
 
 		if resp.StatusCode != http.StatusOK {
-			errMsg := fmt.Sprintf("unable to fetch %s. Return code %d", fullURL, resp.StatusCode)
+			body, err := ioutil.ReadAll(resp.Body)
```
mrnugget: Don't you need to close the `resp.Body`?

api/diagnostics.go
```diff
@@ -519,7 +519,16 @@ func (j *DiagnosticsJob) getHTTPAddToZip(node Node, endpoints map[string]string,
...
+			if err != nil {
+				logrus.WithError(err).Warn("Unable to read response body")
```
mrnugget: Don't you have to close the response body even when `ReadAll` fails? As I understand it, you need to close it whenever you work with the body or _attempt to_, see: https://golang.org/pkg/net/http/

screenshot 2019-02-14 at 09 44 23



How about something like this?

```go
var errMsg string
body, err := ioutil.ReadAll(resp.Body)
resp.Body.Close()
if err != nil {
	errMsg = fmt.Sprintf("unable to fetch %s. Return code %d. Could not read body", fullURL, resp.StatusCode)
} else {
	errMsg = fmt.Sprintf("unable to fetch %s. Return code %d. Body: %s", fullURL, resp.StatusCode, string(body))
}

```

janisz: Thanks

janisz:
Master PR https://github.com/dcos/dcos-diagnostics/pull/88

janisz:
:wave:  @mrnugget  @br-lewis Can you review, please?

PR #88, opened 2019-02-14
Log response body on error

janisz:
Refs: #87


janisz:

PR #90, opened 2019-02-22
Use context instead of custom channel

janisz:
Currently, we are using a custom implementation of context. This change introduces [`standard context`](https://golang.org/pkg/context/) that is supported by HTTP client this it will fix problems with cancelling a running job.

### DIFF
--- a/api/diagnostics.go
+++ b/api/diagnostics.go
@@ -45,7 +45,7 @@ type DiagnosticsJob struct {
 	statusMutex   sync.RWMutex
 	progressMutex sync.RWMutex
 
-	cancelChan   chan bool
+	cancelFunc   context.CancelFunc
 	logProviders logProviders
 	client       *http.Client
 
@@ -156,13 +156,16 @@ func (j *DiagnosticsJob) run(req bundleCreateRequest) (createResponse, error) {
 	t := time.Now()
 	bundleName := fmt.Sprintf("bundle-%d-%02d-%02d-%d.zip", t.Year(), t.Month(), t.Day(), t.Unix())
 
+	ctx, cancelFunc := context.WithTimeout(context.Background(), time.Minute*time.Duration(j.Cfg.FlagDiagnosticsJobTimeoutMinutes))
+
 	j.LastBundlePath = filepath.Join(j.Cfg.FlagDiagnosticsBundleDir, bundleName)
 	j.setStatus("Diagnostics job started, archive will be available at: " + j.LastBundlePath)
-	j.cancelChan = make(chan bool)
+	j.cancelFunc = cancelFunc
 	j.JobStarted = time.Now()
 	j.Running = true
 	j.JobDuration = 0
-	go j.runBackgroundJob(foundNodes)
+	j.JobProgressPercentage = 0
+	go j.runBackgroundJob(ctx, foundNodes)
 
 	var r createResponse
 	r.Extra.LastBundleFile = bundleName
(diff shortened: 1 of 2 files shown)


janisz:

api/diagnostics.go
```diff
@@ -226,7 +208,14 @@ func (j *DiagnosticsJob) runBackgroundJob(nodes []dcos.Node) {
 	// place a summaryErrorsReport.txt in a zip archive which should provide info what failed during the logs collection.
 	summaryErrorsReport := new(bytes.Buffer)
 
-	j.collectDataFromNodes(nodes, summaryReport, summaryErrorsReport, zipWriter)
+	err = j.collectDataFromNodes(ctx, nodes, summaryReport, summaryErrorsReport, zipWriter)
+	j.setJobProgressPercentage(100)
```
mrnugget: I'd move this into a defer in `collectDataFromNodes` since it's the primary setter of the percentages, AFAICT.

api/diagnostics.go
```diff
@@ -249,12 +238,17 @@ func (j *DiagnosticsJob) flushReport(zipWriter *zip.Writer, fileName string, rep
 	}
 }
 
-func (j *DiagnosticsJob) collectDataFromNodes(nodes []dcos.Node, summaryReport *bytes.Buffer,
-	summaryErrorsReport *bytes.Buffer, zipWriter *zip.Writer) {
-	j.setJobProgressPercentage(0)
```
mrnugget: Did this get lost in the process? Nothing sets `0` anymore?

api/diagnostics_test.go
```diff
@@ -409,6 +409,66 @@ func TestGetStatusWhenJobIsRunning(t *testing.T) {
...
+		MR:               &MonitoringResponse{},
+	}
+
+	dt.DtDiagnosticsJob.run(bundleCreateRequest{Nodes: []string{"all"}})
+
+	<-called
+	dt.DtDiagnosticsJob.cancel()
+	wait <- false
+
+	for s := job.getBundleReportStatus(); s.Running; s = job.getBundleReportStatus() {
```
mrnugget: I think you can reduce that to: `for job.getBundleReportStatus().Running { }`

api/diagnostics_test.go
```diff
@@ -409,6 +409,66 @@ func TestGetStatusWhenJobIsRunning(t *testing.T) {
...
+	dt := &Dt{
+		Cfg:              cfg,
+		DtDCOSTools:      tools,
+		DtDiagnosticsJob: job,
+		MR:               &MonitoringResponse{},
+	}
+
+	dt.DtDiagnosticsJob.run(bundleCreateRequest{Nodes: []string{"all"}})
+
+	<-called
```
mrnugget: You could add an assertion below this to make sure that the `getBundleReportStatus` is already `Running` at that point

api/diagnostics.go
janisz: Hmm, I've just take it out but maybe you're right

api/diagnostics.go
janisz: It is set here https://github.com/dcos/dcos-diagnostics/pull/90/files#diff-0ee9de7b9d9927cce9fb53251662d72aR167 I moved it because I thought it will be clear that we set 0 at the beginning and 100 at the end

janisz:
> And, looking at it again: am I correct in assuming that the main reason for this change is to have better timeout handling?

Yes and the second thing is I'd like to extract `collectDataFromNodes` function to a separated package and `context` is more generic than a custom channel.

mrnugget:
> Yes and the second thing is I'd like to extract `collectDataFromNodes` function to a separated package and `context` is more generic than a custom channel.

Okay, gotcha!

PR #91, opened 2019-02-27
Create zip for each file

janisz:
Allow fetching data in parallel

### DIFF
--- a/api/diagnostics.go
+++ b/api/diagnostics.go
@@ -330,7 +330,24 @@ NodeLoop:
 
 	numberOfEndpointsToFetch := len(fetchRequests)
 	percentPerEndpoint := 100.0 / float32(numberOfEndpointsToFetch)
-	done := j.collectStatusUpdates(ctx, numberOfEndpointsToFetch, fetchStatusUpdate, percentPerEndpoint, summaryReport, summaryErrorsReport)
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		for i := 0; i < numberOfEndpointsToFetch; i++ {
+			select {
+			case <-ctx.Done():
+				return
+			case status := <-fetchStatusUpdate:
+				j.incJobProgressPercentage(percentPerEndpoint)
+				e := status.Error
+				updateSummaryReportBuffer("GET "+status.URL, fmt.Sprint(e), summaryReport)
+				j.setStatus("GET " + status.URL)
+				if e != nil {
+					j.logError(e, status.URL, summaryErrorsReport)
+				}
+			}
+		}
+	}()
 
 	for _, r := range fetchRequests {
 		fetchReq <- r
@@ -353,28 +370,6 @@ NodeLoop:
 		return zips, fmt.Errorf("diagnostics job failed: %v", allErrors)
 	}
 	return zips, nil
-}
-
-func (j *DiagnosticsJob) collectStatusUpdates(ctx context.Context, numberOfEndpointsToFetch int, fetchStatusUpdate chan fetcher.FetchStatusUpdate, percentPerEndpoint float32, summaryReport *bytes.Buffer, summaryErrorsReport *bytes.Buffer) <-chan struct{} {
-	done := make(chan struct{})
-	go func() {
-		defer close(done)
-		for i := 0; i < numberOfEndpointsToFetch; i++ {
-			select {
-			case <-ctx.Done():
-				return
-			case status := <-fetchStatusUpdate:
-				j.incJobProgressPercentage(percentPerEndpoint)
-				e := status.Error
-				updateSummaryReportBuffer("GET "+status.URL, fmt.Sprint(e), summaryReport)
-				j.setStatus("GET " + status.URL)
-				if e != nil {
-					j.logError(e, status.URL, summaryErrorsReport)
-				}
-			}
-		}
-	}()
-	return done
 }
 
 func (j *DiagnosticsJob) setJobProgressPercentage(v float32) {
--- a/api/diagnostics.go
+++ b/api/diagnostics.go
@@ -219,13 +219,9 @@ func (j *DiagnosticsJob) runBackgroundJob(ctx context.Context, nodes []dcos.Node
 	j.setJobProgressPercentage(100)
 
 	for _, path := range zips {
-		e := appendToZip(zipWriter, path)
-		if e != nil {
-			j.logError(e, "Could not create a bundle", summaryErrorsReport)
+		if err := appendToZip(zipWriter, path); err != nil {
+			j.logError(err, "Could not create a bundle", summaryErrorsReport)
 		}
-	}
-
-	for _, path := range zips {
 		if err = os.Remove(path); err != nil {
 			j.logError(err, "Could not remove temporary file", summaryErrorsReport)
 		}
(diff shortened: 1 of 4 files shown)


janisz:
:wave:  @mrnugget @drewvanstone Can you review, please?

fetcher/fetcher.go
```diff
@@ -0,0 +1,151 @@
...
+type Fetcher struct {
+	file         *os.File
+	client       *http.Client
+	input        <-chan EndpointFetchRequest
+	statusUpdate chan<- FetchStatusUpdate
+	output       chan<- FetchBulkResponse
+}
+
+// New creates new Fetcher. Fetcher needs to be started with Run()
+func New(
```
drewvanstone: If I understand correctly, it looks like the fetcher object is created just once whenever we need to generate a diagnostics bundle. If that's the case, couldn't this `New()` function just create the channels and http client itself? Ideally objects like these should be able to be created with just `New()`, instead of having a long list of parameters to pass in.

fetcher/fetcher.go
```diff
@@ -0,0 +1,151 @@
...
+	"net/http"
+	"os"
+	"path/filepath"
+
+	"github.com/dcos/dcos-diagnostics/dcos"
+	"github.com/sirupsen/logrus"
+)
+
+// EndpointFetchRequest is a struct passed to Fetcher with information about URL to be fetched
+type EndpointFetchRequest struct {
```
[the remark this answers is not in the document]
drewvanstone: Since this is already the `fetcher` package, how about `EndpointRequest`, so it'll be `fetcher.EndpointRequest`.

fetcher/fetcher.go
```diff
@@ -0,0 +1,151 @@
...
+// FetchBulkResponse is a message published when Fetcher finish it job due to cancelled context or closed input chanel
+type FetchBulkResponse struct {
+	ZipFilePath string
+}
+
+// Fetcher is a struct responsible for fetching nodes endpoints
+type Fetcher struct {
+	file         *os.File
+	client       *http.Client
+	input        <-chan EndpointFetchRequest
```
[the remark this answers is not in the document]
drewvanstone: I agree. I like `endpoints` and `results`

fetcher/fetcher.go
```diff
@@ -0,0 +1,151 @@
...
+
+// Run starts fetcher. This method should be run as a gorutine
+func (f *Fetcher) Run(ctx context.Context) {
+	zipWriter := zip.NewWriter(f.file)
+
+	f.workOffRequests(ctx, zipWriter)
+
+	filename := f.file.Name()
+	if err := zipWriter.Close(); err != nil {
+		logrus.WithError(err).Errorf("Could not close zip writer %s", f.file.Name())
```
drewvanstone: I generally would rather create error messages and bubble them up to the caller, rather than calling them directly deep in a package. If we were to move away from Logrus, for example, we'd have to come into the package to untangle it.

fetcher/fetcher_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"net/url"
+	"testing"
+
+	"github.com/dcos/dcos-diagnostics/dcos"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_NewReturnErrorWhenCantCreateZip(t *testing.T) {
+	_, err := New("not_existing_dir", nil, nil, nil, nil)
```
drewvanstone: So if New() were to create the channels and client, then this call would just be `New("not_existing_dir")`, which is a lot easier to read and write.

fetcher/fetcher.go
mrnugget: Yeah, currently it's only created once, but the idea is to create ~5 of them to download the endpoints concurrently. And then you need to pass the channels in so they can all put their results in the same channel, etc.

fetcher/fetcher.go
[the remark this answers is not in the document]
mrnugget: Nice!

api/diagnostics.go
```diff
@@ -208,7 +209,24 @@ func (j *DiagnosticsJob) runBackgroundJob(ctx context.Context, nodes []dcos.Node
 	// place a summaryErrorsReport.txt in a zip archive which should provide info what failed during the logs collection.
 	summaryErrorsReport := new(bytes.Buffer)
 
-	err = j.collectDataFromNodes(ctx, nodes, summaryReport, summaryErrorsReport, zipWriter)
+	zips, err := j.collectDataFromNodes(ctx, nodes, summaryReport, summaryErrorsReport)
+	defer func(files []string) {
```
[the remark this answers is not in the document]
janisz: `err` is checked below. Fixed

fetcher/fetcher.go
janisz: Do you think we should put `error` field in response struct?

fetcher/fetcher_test.go
janisz: but then we lose control if channels are buffered or not

drewvanstone: Ya, @mrnugget pointed that out above

fetcher/fetcher.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not create temp zip file in %s: %s", tempdir, err)
+	}
+
+	fetcher := &Fetcher{f, client, input, statusUpdate, output}
+
+	return fetcher, nil
+}
+
+// Run starts fetcher. This method should be run as a gorutine
```
drewvanstone: `/gorutine/goroutine`

mrnugget:
btw. did you try running the code with more than 1 worker?

api/diagnostics.go
```diff
@@ -239,43 +280,100 @@ func (j *DiagnosticsJob) flushReport(zipWriter *zip.Writer, fileName string, rep
...
 
-		updateSummaryReportBuffer("START collecting logs", node, "", summaryReport)
-		endpoints, err := j.getNodeEndpoints(node)
+	fetchRequests := j.getEndpointsToFetch(ctx, nodes, summaryReport, summaryErrorsReport)
+
+	fetchReq := make(chan fetcher.EndpointRequest, len(fetchRequests))
+	for _, r := range fetchRequests {
+		fetchReq <- r
+	}
+	close(fetchReq)
```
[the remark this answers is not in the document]
janisz: We need inform workes that tere is no more work to do

janisz:
> btw. did you try running the code with more than 1 worker?
Nope, I'll do it in a minute

api/diagnostics.go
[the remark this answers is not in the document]
mrnugget: TIL that you can close buffered channels before reading from them! (https://www.rapidloop.com/blog/golang-channels-tips-tricks.html)

fetcher/fetcher.go
drewvanstone: Ya that could work

janisz:
@mrnugget 
> btw. did you try running the code with more than 1 worker?

More then one workere test https://github.com/dcos/dcos/pull/4659/commits/c969619c1863bdf73d786c7e0e3fec30cd075a3c

mrnugget:
Ah! And it looks like all the tests are passing, right?

PR #95, opened 2019-04-01
Switch to Go Modules

makkes:
* Removed dep's Gopkg.* files
* Introduced go.mod and go.sum
* Bumped dev Docker image to Go 1.12
* Switched from gometalinter go golangci-lint
* Adapted Makefile (and removed the unnecessary `--privileged` switch)
* Adapted Travis CI configuration

https://jira.mesosphere.com/browse/DCOS_OSS-4984


janisz:
There are problems with gometalinter. I think we can switch to something newer/faster

PR #99, opened 2019-04-09
Allow optional commands

janisz:
Currently if command does not exists we create empty file. This change includes `stderr` in command output file and if command does not exists or error (returns non `0` result) then mark it as failed and note it in `summarryErrorReport.txt`. This change introduce also `Optional` field for command so event if they fail, the bundle could be finished whiteout errors just like files.

Fixes: DCOS_OSS-5026
Refs: https://github.com/dcos/dcos-diagnostics/pull/94

### DIFF
--- a/api/diagnostics.go
+++ b/api/diagnostics.go
@@ -12,6 +12,7 @@ import (
 	"math/rand"
 	"net/http"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"sync"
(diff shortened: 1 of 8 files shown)


makkes:
@janisz I remember that in our grooming you said that there is the possibility that this change breaks other stuff. Can you elaborate on this here or did I just get it wrong in the meeting?

janisz:
I've checked it and new optional command are already used in EE config so this should not break anything

janisz:
But we still has breaking tests in OSS

PR #100, opened 2019-04-10
Standardize Makefile targets; add git revision to version

makkes:
fixes https://jira.mesosphere.com/browse/DCOS_OSS-5008

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,15 @@
 DEFAULT_TARGET: build
+
+VERSION := 0.4.0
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -1,33 +1,46 @@
 DEFAULT_TARGET: build
+
+VERSION := 0.4.0
```
janisz: How we would like to bump this value?

makkes: IMO this is still to be discussed. I just wanted to do a first step in the direction of versioning all of our components in the same way.

PR #101, opened 2019-04-10
Optional http endpoints

jdef:
Notable:
* API change for content returned by `/logs` of the diagnostics component. If this is an internal API to the diagnostics subsystem, and there are no other dependencies other than what's been changed in this PR, then we should be OK. Otherwise ... we'll need to consider an alternative approach here.

Related to https://jira.mesosphere.com/browse/DCOS-51618

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -31,6 +31,7 @@ type EndpointRequest struct {
 	URL      string
 	Node     dcos.Node
 	FileName string
+	Optional bool
 }
 
 // StatusUpdate is an update message published by Fetcher when EndpointRequest is done. If error occurred	 during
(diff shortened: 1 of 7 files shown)


jdef:
@janisz PTAL

jdef:
squashed and attempted to sign

fetcher/fetcher.go
```diff
@@ -122,7 +123,11 @@ func getDataToZip(ctx context.Context, client *http.Client, r EndpointRequest, z
 
 	resp, err := get(ctx, client, r.URL)
 	if err != nil {
-		return fmt.Errorf("could not get from url %s: %s", r.URL, err)
+		if !r.Optional {
+			return fmt.Errorf("could not get from url %s: %s", r.URL, err)
+		}
+		logrus.Infof("Failed to fetch OPTIONAL URL %s: %v", r.URL, err)
```
janisz:
```suggestion
		logrus.WithError(err).Infof("Failed to fetch OPTIONAL URL %s", r.URL)
```

jdef:
re-pushed, all commits signed

jdef:
ugh. GH suggestions aren't signed?! I just signed the suggested change and repushed

PR #102, opened 2019-04-11
Read providers configuration from multiple files

janisz:
This change replaces `FlagDiagnosticsBundleEndpointsConfigFile string` with `FlagDiagnosticsBundleEndpointsConfigFiles []string` to allow configure multiple files.
Files will be merged and last added file will override previous one (order is important). This change allows us to have common configuration between OSS and EE and keep EE specific configuration in separated file.

Fixes: [DCOS-51484](https://jira.mesosphere.com/browse/DCOS-51484)


janisz:

PR #103, opened 2019-04-16
Make number of fetchers configurable

janisz:
Fixes: [DCOS_OSS-4964](https://jira.mesosphere.com/browse/DCOS_OSS-4964)

### DIFF
--- a/api/diagnostics.go
+++ b/api/diagnostics.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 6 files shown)


api/diagnostics.go
```diff
@@ -303,7 +304,7 @@ func (j *DiagnosticsJob) collectDataFromNodes(ctx context.Context, nodes []dcos.
 	fetchStatusUpdate := make(chan fetcher.StatusUpdate)
 	fetchResponse := make(chan fetcher.BulkResponse)
 
-	numberOfWorkers := 1 //TODO(janisz): make number of workers configurable
+	numberOfWorkers := j.Cfg.FlagDiagnosticsBundleFetchersCount
```
makkes: Would you be able to add a unit test that asserts that the number of fetchers created corresponds to the configuration flag?

janisz: Hmm, I'm not sure how to test it. In theory number of workesrs is transparent we could try to catch concurrent requests but I don't think this is easy

makkes: Okay, let's leave this as an exercise for a rainy day...

PR #105, opened 2019-04-24
Added timeout to systemd log reading; fixed `verbose` flag

makkes:
This closes https://jira.mesosphere.com/browse/DCOS_OSS-5097

I added a TimeoutReadCloser to our units package that reads data from
an underlying ReadCloser only when the deadline of the given Context
has not been reached. Otherwise it stops reading and returns the error
retrieved from the Context (usually a DeadlineExceeded error).

When reading a systemd unit's log is interrupted we now log this fact
in the diagnostics output so that the consumer of the bundle knows
there's probably more in the journal than he sees.

### DIFF
--- a/units/helpers.go
+++ b/units/helpers.go
@@ -1,0 +1,22 @@
+package units
+
+import (
+	"context"
+	"io"
+)
+
+type TimeoutReadCloser struct {
+	ctx context.Context
+	src io.ReadCloser
+}
+
+func (r *TimeoutReadCloser) Read(p []byte) (n int, err error) {
+	if err := r.ctx.Err(); err != nil {
+		return 0, err
+	}
+	return r.src.Read(p)
+}
+
+func (r *TimeoutReadCloser) Close() error {
+	return r.src.Close()
+}
--- a/units/helpers_test.go
+++ b/units/helpers_test.go
@@ -1,9 +1,12 @@
 package units
 
 import (
+	"context"
 	"io/ioutil"
 	"runtime"
+	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -14,7 +17,7 @@ func TestReadJournalOutputSince_Windows(t *testing.T) {
 		t.Skip()
 	}
 
-	r, err := ReadJournalOutputSince("", "")
+	r, err := ReadJournalOutputSince(context.TODO(), "", "")
 	assert.Nil(t, r)
 	assert.EqualError(t, err, "there is no journal on Windows")
 }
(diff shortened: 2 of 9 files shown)


units/helpers.go
```diff
@@ -0,0 +1,22 @@
...
+	"io"
+)
+
+type TimeoutReadCloser struct {
+	ctx context.Context
+	src io.ReadCloser
+}
+
+func (r *TimeoutReadCloser) Read(p []byte) (n int, err error) {
+	if err := r.ctx.Err(); err != nil {
```
janisz: I don't understand how it works.
 IMO if context is cancelled but we can read and finish operation we should do it. If reading takes long and context could be cancelled in meantime we should cancel it. This only works when context is cancelled before reading which is not something we are looking for.

units/helpers_test.go
```diff
@@ -24,11 +27,34 @@ func TestReadJournalOutputSince_Linux(t *testing.T) {
...
 	data, err := ioutil.ReadAll(r)
 	require.NoError(t, err)
 
 	assert.Empty(t, data)
+
+	assert.IsType(t, &TimeoutReadCloser{}, r)
+	tr := r.(*TimeoutReadCloser)
+	deadline, ok := tr.ctx.Deadline()
+	assert.True(t, ok, "Context expected to have a deadline but it had none")
+	assert.Equal(t, ctxDeadline, deadline, "Deadline has unexpected value")
```
janisz: What are you testing here?

units/helpers.go
makkes: Let me elaborate on how this works:

* A context is created at time 0 with deadline 5.
* At time 1 `Read` is called and returns the desired amount of bytes (at most).
* At time 2 `Read` is called and returns the desired amount of bytes (at most).
* At time 5 the deadline is reached.
* At time 6 `Read` is called but since the deadline has already been reached we don't read anything but return immediately.

What this does NOT account for is that when during a `Read` operation the deadline is reached. The `Read` operation of the underlying source will NOT be interrupted. But this is IMO an edge case we can probably live with. It's more important to read as much data as possible in the given time frame and stop reading when time's up.

janisz: OK, got it!

units/helpers_test.go
makkes: We want to make sure that `ReadJournalOutputSince` creates a `TimeoutReadCloser` wrapper for the underlying systemd `ReadCloser` with the deadline that we provided when calling `ReadJournalOutputSince`.

units/helpers.go
janisz: Related: https://github.com/golang/go/issues/20280

PR #106, opened 2019-04-25
Remove usage of dcos-history

janisz:
dcos-history is deprecated so we will not use it anymore. It was used to detect nodes in case when DNS is down. I think relying only on DNS should be enough in this service. With dcos-hisotry this patch removes selfTest that was used only to test integration with dcos-history.


api/router.go
```diff
@@ -223,11 +223,6 @@ func getRoutes(dt *Dt) []routeHandler {
 			handler: h.deleteBundleHandler,
 			methods: []string{"POST"},
 		},
-		// self test route
-		{
-			url:     baseRoute + "/selftest/info",
```
janisz: Do you think we should keep this endpoint for backward compatibility even if it returns dummy response?

janisz:
Reverts https://github.com/dcos/dcos-diagnostics/commit/ed1fed97cc1bc3b6d9eaa048cb7b68a68d44a542

PR #107, opened 2019-05-06
Get rid of `fetchHistogram` global variable

makkes:
The histogram is now injected into Fetcher structs so we can also unit
test its calls properly.

refs https://jira.mesosphere.com/browse/DCOS_OSS-5136

### DIFF
--- a/api/diagnostics.go
+++ b/api/diagnostics.go
@@ -67,7 +67,8 @@ type DiagnosticsJob struct {
 	JobEnded              time.Time
 	JobDuration           time.Duration
 	JobProgressPercentage float32
-	fetchVec              prometheus.ObserverVec
+	// This vector is used to collect the HTTP response times of all endpoints.
+	FetchPrometheusVector prometheus.ObserverVec
 }
 
 type logProviders struct {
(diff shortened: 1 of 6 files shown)


api/diagnostics.go
```diff
@@ -67,6 +67,7 @@ type DiagnosticsJob struct {
 	JobEnded              time.Time
 	JobDuration           time.Duration
 	JobProgressPercentage float32
+	fetchVec              prometheus.ObserverVec
```
janisz: I think the name could be more descriptive and/or type could be more precise. What exactly we will collect there?

fetcher/fetcher.go
```diff
@@ -50,11 +41,12 @@ type BulkResponse struct {
...
-	client       *http.Client
-	endpoints    <-chan EndpointRequest
-	statusUpdate chan<- StatusUpdate
-	results      chan<- BulkResponse
+	file          *os.File
+	client        *http.Client
+	endpoints     <-chan EndpointRequest
+	statusUpdate  chan<- StatusUpdate
+	results       chan<- BulkResponse
+	prometheusVec prometheus.ObserverVec
```
janisz: I think the name could be more descriptive and/or type could be more precise. What exactly we will collect there?

fetcher/fetcher.go
```diff
@@ -13,20 +13,11 @@ import (
...
-	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promauto"
-
 	"github.com/dcos/dcos-diagnostics/dcos"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
 )
 
-var (
-	fetchHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
```
janisz: Where the histogram is created now?

PR #112, opened 2019-05-27
Add CRUD like API for diagnostics bundles

janisz:
The API definition could be found [here](https://janisz.docs.apiary.io/#reference/0/diagnostics-bundles)

### DIFF
--- a/api/providers.go
+++ b/api/providers.go
@@ -142,7 +142,7 @@ func LoadCollectors(cfg *config.Config, tools dcos.Tooler, client *http.Client) 
 			fileName = endpoint.FileName
 		}
 		//TODO(janisz): Handle socket-only endpoints e.g., dcos-diagnostics
-		url, err := util.UseTLSScheme(fmt.Sprintf("http://localhost:%d%s", endpoint.Port, endpoint.URI), cfg.FlagForceTLS)
+		url, err := util.UseTLSScheme(fmt.Sprintf("http://%s:%d%s", cfg.FlagHostname, endpoint.Port, endpoint.URI), cfg.FlagForceTLS)
 		if err != nil {
 			return nil, fmt.Errorf("could not initialize internal log providers: %s", err)
 
--- a/api/providers.go
+++ b/api/providers.go
@@ -141,7 +141,7 @@ func LoadCollectors(cfg *config.Config, tools dcos.Tooler, client *http.Client) 
 		if endpoint.FileName != "" {
 			fileName = endpoint.FileName
 		}
-		//TODO(janisz): Handle socket-only endpoints e.g., dcos-diagnostics
+
 		url, err := util.UseTLSScheme(fmt.Sprintf("http://%s:%d%s", cfg.FlagHostname, endpoint.Port, endpoint.URI), cfg.FlagForceTLS)
 		if err != nil {
 			return nil, fmt.Errorf("could not initialize internal log providers: %s", err)
@@ -177,7 +177,7 @@ func LoadCollectors(cfg *config.Config, tools dcos.Tooler, client *http.Client) 
 
 	}
 
-	//TODO(janisz): Setup systemd units
+	//TODO(janisz): Setup systemd units https://jira.mesosphere.com/browse/DCOS_OSS-5223
 
 	return coll, nil
 }
--- a/api/providers.go
+++ b/api/providers.go
@@ -4,6 +4,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"github.com/dcos/dcos-diagnostics/collectors"
+	"github.com/dcos/dcos-diagnostics/util"
 
 	"github.com/dcos/dcos-diagnostics/config"
 	"github.com/dcos/dcos-diagnostics/dcos"
@@ -112,4 +117,86 @@ func loadInternalProviders(cfg *config.Config, DCOSTools dcos.Tooler) (internalC
 	return LogProviders{
 		HTTPEndpoints: httpEndpoints,
 	}, nil
+}
+
+func LoadCollectors(cfg *config.Config, tools dcos.Tooler, client *http.Client) ([]collectors.Collector, error) {
+	providers, err := loadProviders(cfg, tools)
+	if err != nil {
+		return nil, fmt.Errorf("could not init bundle handler: %s", err)
+	}
+	var coll []collectors.Collector
+
+	role, err := tools.GetNodeRole()
+	if err != nil {
+		return nil, fmt.Errorf("could not get role: %s", err)
+	}
+
+	// set filename if not set, some endpoints might be named e.g., after corresponding unit
+	for _, endpoint := range providers.HTTPEndpoints {
+		if !roleMatched(role, endpoint.Role) {
+			continue
+		}
+
+		fileName := fmt.Sprintf("%d-%s.json", endpoint.Port, util.SanitizeString(endpoint.URI))
+		if endpoint.FileName != "" {
+			fileName = endpoint.FileName
+		}
+		//TODO(janisz): Handle socket-only endpoints e.g., dcos-diagnostics
+		url, err := util.UseTLSScheme(fmt.Sprintf("http://localhost:%d%s", endpoint.Port, endpoint.URI), cfg.FlagForceTLS)
+		if err != nil {
+			return nil, fmt.Errorf("could not initialize internal log providers: %s", err)
+
+		}
+
+		c := collectors.EndpointCollector{
+			NameOptional: collectors.NameOptional{
+				Name:     fileName,
+				Optional: endpoint.Optional,
+			},
+			URL:    url,
+			Client: client, //TODO(janisz): Check if we can use default client
+		}
+		coll = append(coll, c)
+	}
+
+	// trim left "/" and replace all slashes with underscores.
+	for _, fileProvider := range providers.LocalFiles {
+		if !roleMatched(role, fileProvider.Role) {
+			continue
+		}
+
+		key := strings.Replace(strings.TrimLeft(fileProvider.Location, "/"), "/", "_", -1)
+		c := collectors.FileCollector{
+			NameOptional: collectors.NameOptional{
+				Name:     key,
+				Optional: fileProvider.Optional,
+			},
+			FilePath: fileProvider.Location,
+		}
+		coll = append(coll, c)
+	}
+
+	// sanitize command to use as filename
+	for _, commandProvider := range providers.LocalCommands {
+		if !roleMatched(role, commandProvider.Role) {
+			continue
+		}
+
+		cmdWithArgs := strings.Join(commandProvider.Command, "_")
+		trimmedCmdWithArgs := strings.Replace(cmdWithArgs, "/", "", -1)
+		key := fmt.Sprintf("%s.output", trimmedCmdWithArgs)
+		c := collectors.CmdCollector{
+			NameOptional: collectors.NameOptional{
+				Name:     key,
+				Optional: commandProvider.Optional,
+			},
+			Cmd: commandProvider.Command,
+		}
+		coll = append(coll, c)
+
+	}
+
+	//TODO(janisz): Setup systemd units
+
+	return coll, nil
 }
--- a/api/providers.go
+++ b/api/providers.go
@@ -148,14 +148,7 @@ func LoadCollectors(cfg *config.Config, tools dcos.Tooler, client *http.Client) 
 
 		}
 
-		c := collectors.EndpointCollector{
-			NameOptional: collectors.NameOptional{
-				Name:     fileName,
-				Optional: endpoint.Optional,
-			},
-			URL:    url,
-			Client: client, //TODO(janisz): Check if we can use default client
-		}
+		c := collectors.NewEndpointCollector(fileName, endpoint.Optional, url, client)
 		coll = append(coll, c)
 	}
 
@@ -166,13 +159,7 @@ func LoadCollectors(cfg *config.Config, tools dcos.Tooler, client *http.Client) 
 		}
 
 		key := strings.Replace(strings.TrimLeft(fileProvider.Location, "/"), "/", "_", -1)
-		c := collectors.FileCollector{
-			NameOptional: collectors.NameOptional{
-				Name:     key,
-				Optional: fileProvider.Optional,
-			},
-			FilePath: fileProvider.Location,
-		}
+		c := collectors.NewFileCollector(key, fileProvider.Optional, fileProvider.Location)
 		coll = append(coll, c)
 	}
 
@@ -185,13 +172,7 @@ func LoadCollectors(cfg *config.Config, tools dcos.Tooler, client *http.Client) 
 		cmdWithArgs := strings.Join(commandProvider.Command, "_")
 		trimmedCmdWithArgs := strings.Replace(cmdWithArgs, "/", "", -1)
 		key := fmt.Sprintf("%s.output", trimmedCmdWithArgs)
-		c := collectors.CmdCollector{
-			NameOptional: collectors.NameOptional{
-				Name:     key,
-				Optional: commandProvider.Optional,
-			},
-			Cmd: commandProvider.Command,
-		}
+		c := collectors.NewCmdCollector(key, commandProvider.Optional, commandProvider.Command)
 		coll = append(coll, c)
 
 	}
--- a/api/providers.go
+++ b/api/providers.go
@@ -7,7 +7,7 @@ import (
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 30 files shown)


api/rest/bundle_handler.go
```diff
@@ -0,0 +1,354 @@
...
+	bundle := Bundle{
+		ID:      id,
+		Started: h.clock.Now(),
+		Status:  Started,
+	}
+
+	bundleStatus := jsonMarshal(bundle)
+	err = ioutil.WriteFile(stateFilePath, bundleStatus, filePerm)
+	if err != nil {
+		writeJSONError(w, http.StatusInsufficientStorage, fmt.Errorf("could not update state file %s: %s", id, err))
```
makkes: same as above.

api/rest/bundle_handler.go
```diff
@@ -0,0 +1,354 @@
...
+		return
+	}
+
+	dataFile, err := os.Create(filepath.Join(h.workDir, id, dataFileName))
+	if err != nil {
+		writeJSONError(w, http.StatusInsufficientStorage, fmt.Errorf("could not create data file %s: %s", id, err))
```
makkes: same as above.

api/rest/bundle_handler.go
```diff
@@ -0,0 +1,354 @@
...
+	}
+
+	ctx := context.Background() //TODO(janisz): Use context with deadline
```
makkes: Let's use `context.WithDeadline` already so we don't run into strange bugs later on.

api/rest/bundle_handler.go
```diff
@@ -0,0 +1,354 @@
...
+
+	go func() {
+		select {
+		case <-ctx.Done():
+			break
+		case bundle.Errors = <-done:
+			bundle.Status = Done
+			bundle.Stopped = h.clock.Now()
+			newBundleStatus := jsonMarshal(bundle)
+			e := ioutil.WriteFile(stateFilePath, newBundleStatus, filePerm)
```
makkes: you can just inline the `jsonMarshal(bundle)` call here and get rid of `newBundleStatus`.

api/rest/bundle_handler.go
```diff
@@ -0,0 +1,354 @@
...
+	vars := mux.Vars(r)
+	id := vars["uuid"]
+
+	http.ServeFile(w, r, filepath.Join(h.workDir, id, dataFileName))
+}
+
+func (h BundleHandler) List(w http.ResponseWriter, r *http.Request) {
+	ids, err := ioutil.ReadDir(h.workDir)
+	if err != nil {
+		writeJSONError(w, http.StatusInsufficientStorage, fmt.Errorf("could not read work dir: %s", err))
```
makkes: same as above, 500 is the better choice since we don't know the cause for this error.

api/rest/bundle_handler.go
```diff
@@ -0,0 +1,354 @@
...
+		if err != nil {
+			logrus.WithField("ID", id.Name()).WithError(err).Warn("There is a problem with bundle")
+		}
+		bundles = append(bundles, bundle)
+
+	}
+
+	body := jsonMarshal(bundles)
+
+	write(w, body)
```
makkes: you can just inline the `jsonMarshal(bundles)` call here and get rid of `body`.

api/rest/bundle_handler.go
```diff
@@ -0,0 +1,354 @@
...
+		return
+	}
+
+	if bundle.Status == Deleted || bundle.Status == Canceled {
+		w.WriteHeader(http.StatusNotModified)
+		write(w, jsonMarshal(bundle))
+		return
+	}
+
+	//TODO(janisz): Handle Canceled Status
```
makkes: Is that something we need to track in a JIRA ticket? Sounds pretty important to me for the functionality of this. What if we released without it? Also: Aren't we handling the `Canceled` state in line 301?

api/router.go
```diff
@@ -13,8 +13,18 @@ import (
 )
 
 // baseRoute a base dcos-diagnostics endpoint location.
+// See: https://github.com/dcos/dcos/blob/5108a310d33084e852b6a9d6f122c02b861e4b97/packages/adminrouter/extra/src/docs/api/nginx.agent.yaml#L56-L62
```
makkes: What does this comment mean? Could you expand it and explain why it's there?

api/router.go
```diff
@@ -13,8 +13,18 @@ import (
...
 const baseRoute string = "/system/health/v1"
 
+// Edpoint for listing all bundles
```
makkes: s/Edpoint/Endpoint/

api/structures.go
```diff
@@ -73,6 +74,7 @@ type Dt struct {
 	Cfg               *config.Config
 	DtDCOSTools       dcos.Tooler
 	DtDiagnosticsJob  *DiagnosticsJob
+	BundleHanlder     rest.BundleHandler
```
makkes: s/BundleHanlder/BundleHandler/

apiary.apib
```diff
@@ -0,0 +1,193 @@
+FORMAT: 1A
```
makkes: Shouldn't we document the (known) 5xx cases here, too?

collectors/collector.go
```diff
@@ -0,0 +1,165 @@
...
+	"os"
+	"os/exec"
+	"time"
+)
+
+// Collector is the interface to abstract data collection from different sources
+type Collector interface {
+	// Name returns the Name of this collector
+	Name() string
+	// Optional returns true if Collector is not mandatory and fails should be ignored
```
makkes: s/fails/failures/

collectors/collector.go
```diff
@@ -0,0 +1,165 @@
...
+	Optional() bool
+	// Collect returns collected data
+	Collect(ctx context.Context) (io.ReadCloser, error)
+}
+
+type NameOptional struct {
```
makkes: Let's not do this, please. It saves us some lines of code but comes with the cost of decreased readibility. Let's just add `Name` and `Optional` explicitly to the structs used as `Collectors`.

collectors/collector.go
```diff
@@ -0,0 +1,165 @@
...
+	Name     string
+	Optional bool
+}
+
+type CmdCollector struct {
```
makkes: Please document this struct, too.

collectors/collector.go
```diff
@@ -0,0 +1,165 @@
...
+	return c.NameOptional.Optional
+}
+
+func (c CmdCollector) Collect(ctx context.Context) (io.ReadCloser, error) {
+	cmd := exec.CommandContext(ctx, c.Cmd[0], c.Cmd[1:]...)
+	output, err := cmd.CombinedOutput()
+	return ioutil.NopCloser(bytes.NewReader(output)), err
+}
+
+// TODO(janisz): Make use of this code instead of calling dcos-diagnostics for units data
```
makkes: I'm wondering: Shall we compose a list of all the TODOs you're adding with this PR and create JIRA tasks for the most important ones? Otherwise I suspect these will stay here forever.

api/rest/bundle_handler.go
```diff
@@ -0,0 +1,354 @@
...
+	_, err := h.getBundleState(id)
+	if err == nil {
+		writeJSONError(w, http.StatusConflict, fmt.Errorf("bundle %s already exists", id))
+		return
+	}
+
+	bundleWorkDir := filepath.Join(h.workDir, id)
+	err = os.MkdirAll(bundleWorkDir, dirPerm)
+	if err != nil {
+		writeJSONError(w, http.StatusInsufficientStorage, fmt.Errorf("could not Create bundle %s workdir: %s", id, err))
```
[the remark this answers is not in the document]
janisz: I used 507 to distinguishes workdir/permisions/storage configuration problems from other application related issues

> The method could not be performed on the resource because the server is unable to store the representation needed to successfully complete the request.

api/rest/bundle_handler.go
janisz: what timeout should I set up here? Should it be configurable? Should I use on of existing timeouts?

collectors/collector.go
janisz: I added it to prevent name clash on `Name()` function and `Name` field. Can you propose new "names" for that `Names`?

api/providers.go
```diff
@@ -113,3 +118,85 @@ func loadInternalProviders(cfg *config.Config, DCOSTools dcos.Tooler) (internalC
...
+	for _, endpoint := range providers.HTTPEndpoints {
+		if !roleMatched(role, endpoint.Role) {
+			continue
+		}
+
+		fileName := fmt.Sprintf("%d-%s.json", endpoint.Port, util.SanitizeString(endpoint.URI))
+		if endpoint.FileName != "" {
+			fileName = endpoint.FileName
+		}
+		//TODO(janisz): Handle socket-only endpoints e.g., dcos-diagnostics
```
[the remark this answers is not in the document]
janisz: It's only for DC/OS EE where some services are hidden behind adminrouter and use sockets for communication.

collectors/collector.go
makkes: Can't you just use lowercase names for the fields?

api/providers.go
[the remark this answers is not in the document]
makkes: But how did we handle that before?

api/rest/bundle_handler.go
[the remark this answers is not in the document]
makkes: But is that something we really need? Would the client behave differently than when it receives a simple 500? If not, just use 500. And we should log this fact in the diagnostics logs.

collectors/collector.go
janisz: Done

collectors/collector.go
janisz: How we will initialize them? Should we have a factory method for all collectors?

api/rest/bundle_handler.go
janisz: Done

api/rest/bundle_handler.go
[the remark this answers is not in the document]
janisz: Client could behave differently. This will reduce customer issues if we can display clear message when CLI gets 507 then just generic message when there is 500.

api/providers.go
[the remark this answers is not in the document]
janisz: We goes thru admin router. We can do it the same right now but there is no point in doing it because we want be independent from adminrouter problems.

cmd/daemon.go
```diff
@@ -161,11 +114,21 @@ func startDiagnosticsDaemon() {
...
+
+	bundleTimeout := time.Minute * time.Duration(defaultConfig.FlagDiagnosticsJobTimeoutMinutes)
+	bundleHandler := rest.NewBundleHandler(defaultConfig.FlagDiagnosticsBundleDir, collectors, bundleTimeout)
+
 	// Inject dependencies used for running dcos-diagnostics.
 	dt := &api.Dt{
 		Cfg:               defaultConfig,
 		DtDCOSTools:       DCOSTools,
 		DtDiagnosticsJob:  diagnosticsJob,
+		BundleHanlder:     bundleHandler,
```
br-lewis: `BundleHanlder` -> `BundleHandler`

api/providers.go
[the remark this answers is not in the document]
makkes: Understood. So as long as we don't handle socket-only endpoints this is not usable in production, right?

collectors/collector.go
makkes: I actually like the Go pattern of having `NewXYZ` functions. So yes, we could do it.

drewvanstone: FYI - Here's the recommendation in [Effective Go](https://golang.org/doc/effective_go.html#composite_literals)

api/providers.go
```diff
@@ -4,6 +4,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"github.com/dcos/dcos-diagnostics/collectors"
+	"github.com/dcos/dcos-diagnostics/util"
```
drewvanstone: Not that we should change this now, but I do want to share this advice: https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

I'd like to see us stay away from these types of package names in the future. I'd add 'tools' to the list of bad names as well.

api/providers.go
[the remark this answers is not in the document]
janisz: It's working on OSS and on EE it needs to go thru adminrouter so configured hostname should be set to external ip.

api/rest/bundle_handler.go
janisz: We partially handle it there. We just do not cancel already cancelled job but there is no code for canceling job and removing created file.

apiary.apib
janisz: Done

api/providers.go
```diff
@@ -4,6 +4,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"github.com/dcos/dcos-diagnostics/collectors"
```
drewvanstone: I think this should be singular: `collector`

api/providers.go
```diff
@@ -113,3 +118,85 @@ func loadInternalProviders(cfg *config.Config, DCOSTools dcos.Tooler) (internalC
 		HTTPEndpoints: httpEndpoints,
 	}, nil
 }
+
+func LoadCollectors(cfg *config.Config, tools dcos.Tooler, client *http.Client) ([]collectors.Collector, error) {
+	providers, err := loadProviders(cfg, tools)
+	if err != nil {
+		return nil, fmt.Errorf("could not init bundle handler: %s", err)
+	}
+	var coll []collectors.Collector
```
drewvanstone: since this is a longer function, I think we should use a more descriptive name, like `collectors`

api/rest/bundle_handler.go
```diff
@@ -0,0 +1,357 @@
...
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/dcos/dcos-diagnostics/collectors"
+
+	"github.com/gorilla/mux"
+
```
drewvanstone: We can remove this newline

api/rest/bundle_handler.go
```diff
@@ -0,0 +1,357 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// work dir contains only directories, each dir is created for single bundle (id is its name) and should contains:
```
drewvanstone: /should contains/should contain/

api/rest/bundle_handler.go
```diff
@@ -0,0 +1,357 @@
...
+	bundles := make([]Bundle, 0, len(ids))
+
+	for _, id := range ids {
+		if !id.IsDir() {
+			continue
+		}
+
+		bundle, err := h.getBundleState(id.Name())
+		if err != nil {
+			logrus.WithField("ID", id.Name()).WithError(err).Warn("There is a problem with bundle")
```
drewvanstone: /problem with bundle/problem with the bundle/

janisz:

api/rest/bundle_handler.go
```diff
@@ -0,0 +1,350 @@
...
+	collectors            []collector.Collector // information what should be in the bundle
+	bundleCreationTimeout time.Duration         // limits how long bundle creation could take
+}
+
+func (h BundleHandler) Create(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	id := vars["id"]
+
+	_, err := h.getBundleState(id)
+	if err == nil {
```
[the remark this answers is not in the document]
janisz: Nope, read code below. We check if bundle already exists and if yes we do not create new.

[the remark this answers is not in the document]
drewvanstone: Hmm I feel like that's a pretty confusing API. So if `getBundleState()` does not return an error, that means a bundle already exists? Maybe we should create a small method like `BundleExists()` that we can call that returns true if it already exists.

api/rest/bundle_handler.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	return nil
+}
+
+func (h BundleHandler) Get(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	id := vars["id"]
+
+	bundle, err := h.getBundleState(id)
```
drewvanstone: Same as comment before. I think a separate `BundleExists(id)` method would make this much clearer.

api/rest/bundle_handler.go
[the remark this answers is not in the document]
janisz: OK, I fixed it. Now we can distinguish 404 error when bundle does not exists from 500 when bundle exists but for some reason we can't read it.

collectors/collector.go
janisz: Done

PR #113, opened 2019-05-31
Add systemd collector

janisz:
Instead of using endpoint collector to call itself we should use direct
systemd collector to simplify process of getting logs and reduce systems
we touch.

Fixes: https://jira.mesosphere.com/browse/DCOS_OSS-5223


api/providers.go
```diff
@@ -81,6 +82,31 @@ func loadExternalProviders(endpointsConfgFiles []string) (externalProviders LogP
...
+	units, err := DCOSTools.GetUnitNames()
+	if err != nil {
+		return nil, fmt.Errorf("could not get unit names: %s", err)
+	}
+
+	units = append(units, cfg.SystemdUnits...)
+	collectors := make([]collector.Collector, 0, len(units))
+
+	sinceString := cfg.FlagDiagnosticsBundleUnitsLogsSinceString
+	duration, err := time.ParseDuration(sinceString)
```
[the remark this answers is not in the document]
janisz: Fixed

api/diagnostics.go
```diff
@@ -973,7 +973,12 @@ func (j *DiagnosticsJob) dispatchLogs(ctx context.Context, provider, entity stri
 			return r, errors.New("Only DC/OS systemd units are available")
 		}
 		logrus.Debugf("dispatching a Unit %s", entity)
-		return units.ReadJournalOutputSince(ctx, entity, j.Cfg.FlagDiagnosticsBundleUnitsLogsSinceString)
+		sinceString := j.Cfg.FlagDiagnosticsBundleUnitsLogsSinceString
+		duration, err := time.ParseDuration(sinceString)
```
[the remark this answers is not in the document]

janisz:

collector/collector_linux_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSystemd_Collect(t *testing.T) {
+	if os.Getenv("TRAVIS") != "" {
+		t.Skipf("SKIPPING: We can not read from journal in Travis")
+	}
+	if !journal.Enabled() {
+		t.Skipf("SKIPPING: Journal not enabled")
```
makkes: Hmm, shouldn't we fail the test here? How do we make sure that the test is actually run somewhere?

collector/collector_linux_test.go
```diff
@@ -0,0 +1,41 @@
...
+	err := journal.Send("test message", journal.PriInfo, map[string]string{"UNIT": "test-unit"})
+	require.NoError(t, err)
+
+	c := NewSystemd(
+		"test",
+		false,
+		"test-unit",
+		time.Second,
+	)
+	r, err := c.Collect(context.TODO())
```
makkes: This might introduce a race condition if the message is not picked up fast enough by journald, right?

collector/collector_linux_test.go
janisz: https://github.com/dcos/dcos-diagnostics/pull/116

collector/collector_linux_test.go
janisz: Right but I think it's more predictable then relying on `systemd-journal.service` to log something

PR #116, opened 2019-06-03
Do NOT skip tests when systemd is disabled

janisz:


janisz:
I'm not sure if we can merge it. It's failing on jenkins
```
journal error: could not connect to journald socket

--- FAIL: TestSystemd_Collect (0.00s)

    require.go:794: 

        	Error Trace:	collector_linux_test.go:22

        	Error:      	Received unexpected error:

        	            	journal error: could not connect to journald socket

        	Test:       	TestSystemd_Collect

=== RUN   TestCmdIsCollector
```

makkes:
> I'm not sure if we can merge it. It's failing on jenkins
> 
> ```
> journal error: could not connect to journald socket
> 
> --- FAIL: TestSystemd_Collect (0.00s)
> 
>     require.go:794: 
> 
>         	Error Trace:	collector_linux_test.go:22
> 
>         	Error:      	Received unexpected error:
> 
>         	            	journal error: could not connect to journald socket
> 
>         	Test:       	TestSystemd_Collect
> 
> === RUN   TestCmdIsCollector
> ```

Hmm, so probably journald is not running on Jenkins? But then this test would not run on Jenkins and not on Travis so where would it actually run?

makkes:
This sounds to me a lot like we need to be able to mock away journald from the systemd collector.

janisz:
I'm against mocking systemd. Better fix our CI to run this tests than mock important part of our system. We have collectors to mock interactions with system. Mocking systemd will be even worse than not having test at all, it will give us false perception everything is right while it could not be working at all.

makkes:
> I'm against mocking systemd. Better fix our CI to run this tests than mock important part of our system. We have collectors to mock interactions with system. Mocking systemd will be even worse than not having test at all, it will give us false perception everything is right while it could not be working at all.

I was not talking about mocking systemd but rather the parts that access systemd and injecting them in a DI-manner so that we can mock those in unit tests. We should also make a better distinction between unit tests and integration tests.

janisz:
Closing since this PR is failing on jenkins

PR #117, opened 2019-06-05
WIP: Leverage new Diagnostics API to speed up bundle creation

br-lewis:


fetcher/bundle.go
```diff
@@ -0,0 +1,206 @@
...
+)
+
+// BundleRequest represents a request for a node's diagnostic bundle
+type BundleRequest struct {
+	BundleID string
+	Node     *dcos.Node
+	Client   *http.Client
+	port     int
+	forceTLS bool
+	Data     []byte
```
janisz: This are response data so it means `BundleRequests` is also `BundleResponse`. Weird, I'd like to have separated structure for that.

fetcher/bundle.go
```diff
@@ -0,0 +1,206 @@
...
+	request.WithContext(ctx)
+
+	resp, err := b.Client.Do(request)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusOK {
+		body, err := ioutil.ReadAll(resp.Body)
```
janisz: Storing bundle in memory is not the best idea. Bundle from local machine could be pretty big. I think we should write it on disk and then merge into a single bundle like we did in fetcher.go

fetcher/bundle.go
```diff
@@ -0,0 +1,206 @@
...
+			}
+		default:
+			status := b.checkStatus(ctx)
+			if status.Finished {
+				logrus.Infof("%s bundle finished", b.Node.IP)
+				ticker.Stop()
+
+				// only send an update when finished
+				statusListener <- status
+			}
```
janisz:
```suggestion
				return
			}
```

api/diagnostics.go
```diff
@@ -298,118 +258,125 @@ func (j *DiagnosticsJob) flushReport(zipWriter *zip.Writer, fileName string, rep
...
-	fetchResponse := make(chan fetcher.BulkResponse)
+	requests := j.getBundleRequests(bundleID, nodes, summaryReport)
+	requestFinished := make(chan fetcher.RequestStatus)
 
-	numberOfWorkers := j.Cfg.FlagDiagnosticsBundleFetchersCount
-	for i := 0; i < numberOfWorkers; i++ {
-		f, err := fetcher.New(j.Cfg.FlagDiagnosticsBundleDir, j.client, fetchReq, fetchStatusUpdate, fetchResponse, j.FetchPrometheusVector)
+	// send creation requests to all nodes
+	for _, r := range requests {
+		err := r.SendCreationRequest(ctx, requestFinished)
```
makkes: we should probably make this asynchronous in the same way as the `downloadBundle` call is asynchronous.

Also, `SendCreationRequest` should always send sth. to `requestFinished` and not return anything so that the number of items in the channel is equal to the number of requests.

janisz:
I thought we will support both APIs for a while, just to make sure everything works until we officially release it.

api/diagnostics.go
```diff
@@ -298,118 +258,125 @@ func (j *DiagnosticsJob) flushReport(zipWriter *zip.Writer, fileName string, rep
...
-	zips, errs := gatherAllResults(fetchResponse, numberOfWorkers)
-
-	if len(errs) != 0 {
-		j.logError(fmt.Errorf("%v", errs), "failed to gather all results", summaryErrorsReport)
+	// wait until all requests have reported that they're finished
+	for i := 0; i < len(requests); i++ {
+		select {
+		case <-ctx.Done():
+			return
+		case update := <-requestFinished:
```
makkes: Here we can download the bundle directyle and send it to the zipQueue (i.e. call `downloadBundle`)

api/diagnostics.go
```diff
@@ -172,7 +171,8 @@ func (j *DiagnosticsJob) run(req bundleCreateRequest) (createResponse, error) {
 	j.Errors = nil
 
 	t := time.Now()
-	bundleName := fmt.Sprintf("bundle-%d-%02d-%02d-%d.zip", t.Year(), t.Month(), t.Day(), t.Unix())
+	bundleID := fmt.Sprintf("bundle-%d-%02d-%02d-%d", t.Year(), t.Month(), t.Day(), t.Unix())
```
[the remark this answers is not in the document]
janisz: It can be any unique string. UUID contains info about node and time but date is also fine
http://www.rfcreader.com/#rfc4122_line410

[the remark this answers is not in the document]
makkes: Well, the design doc specifically mentions UUID. That's why I'm nitpicking here.

PR #118, opened 2019-06-10
Do not return end time when job is running.

janisz:
Use empty string instead of random valued of `JobEnded` to prevent user misinformation when job is still running.
Remove `JobDurtion` from diagnosticsJob – it could be calculated in `getBundleReportStatus()`.

Fixes: https://jira.mesosphere.com/browse/DCOS_OSS-5254


api/diagnostics.go
```diff
@@ -620,14 +619,20 @@ func (j *DiagnosticsJob) getBundleReportStatus() bundleReportStatus {
...
-		Running:               j.Running,
+		Running:               running,
 		Status:                stat,
 		Errors:                errors,
 		LastBundlePath:        j.LastBundlePath,
 		JobStarted:            j.JobStarted.String(),
-		JobEnded:              j.JobEnded.String(),
-		JobDuration:           j.JobDuration.String(),
+		JobEnded:              ended,
+		JobDuration:           time.Since(j.JobStarted).String(),
```
drewvanstone: If a job has completed, wouldn't duration equal `JobEnded - JobStarted`?

api/diagnostics_test.go
```diff
@@ -567,6 +567,7 @@ func TestCancelWhenJobIsRunning(t *testing.T) {
 
 	<-called
 	require.True(t, job.getBundleReportStatus().Running)
+	require.Empty(t, job.getBundleReportStatus().JobEnded, "job is running, end time should be empty")
```
drewvanstone: Personally, I'm not really a fan of using empty/non-empty strings for boolean logic. Shouldn't we just check if the job is running or not using `job.Running` like you do on the line above?

janisz: This is test to check if logic above is right

api/diagnostics.go

api/diagnostics_test.go
drewvanstone: Ah, makes sense.

PR #120, opened 2019-06-10
Remove dependency to dcos-log

janisz:
Use github.com/coreos/go-systemd directly without dcos-log wrapper.

Fixes: https://jira.mesosphere.com/browse/DCOS_OSS-5261


janisz:
@makkes I created issue for this https://jira.mesosphere.com/browse/DCOS_OSS-5261

PR #121, opened 2019-06-11
Add test assertions

makkes:
This seems to mitigate https://jira.mesosphere.com/browse/DCOS_OSS-5241


drewvanstone:
> This seems to mitigate https://jira.mesosphere.com/browse/DCOS_OSS-5241

How does this fix it? Not sure I'm following. The mock sometimes has unset and thus nil values, and this `AssertExpectations` just makes sure everything is set, is that right?

janisz:
Jenkins failed. I think we can lower the assertion [here](https://github.com/dcos/dcos-diagnostics/blob/2457ed661d38f65fedede99ab0ec8e5b372cdc4f/collector/collector_test.go#L165) and only check if error occured

``` diff
collector_test.go:165: 
Error Trace:	collector_test.go:165
  Error:      	Error message not equal:
-    expected: "could not fetch url invalid url: Get invalid%20url: unsupported protocol scheme \"\""
+    actual  : "could not fetch url invalid url: Get invalid%20url: unsupported protocol scheme \"\" (Client.Timeout exceeded while awaiting headers)"
```

makkes:
```How does this fix it? Not sure I'm following. The mock sometimes has unset and thus nil values, and this AssertExpectations just makes sure everything is set, is that right?```

Well, the error is pretty surely caused by some race condition that occurs when running multiple tests. I experienced a lesser number of occurrences of the flake when actually asserting the mocked requests. Also, when mocking sth. you should definitely ascertain that it's used, otherwise you wouldn't need to mock it in the first place.

Since the checks are still failing I'll have a look at this when I'm at the computer again.

janisz:
@makkes Can we merge it?

PR #124, opened 2019-06-17
Add functionality to create a remote bundle

br-lewis:


api/rest/bundle_handler.go
```diff
@@ -69,7 +77,20 @@ type BundleHandler struct {
...
+}
+
+type createArguments struct {
+	BundleType string `json:"type"`
+	Nodes      []node `json:"nodes"`
+}
+
+type node struct {
+	IP      net.IP `json:"ip"`
+	Role    string `json:"role"`
```
janisz: When do we serialize `node` to JSON?

br-lewis: We serialize it from json. This is a change to the API in order to get the nodes' roles though maybe we could pull this from within the cluster using the IP alone.

api/rest/bundle_handler.go
```diff
@@ -69,7 +77,20 @@ type BundleHandler struct {
...
 	workDir               string                // location where bundles are generated and stored
 	collectors            []collector.Collector // information what should be in the bundle
+	coordinator           Coordinator           // coordinator for building remote bundles
 	bundleCreationTimeout time.Duration         // limits how long bundle creation could take
+	urlBuilder            dcos.NodeURLBuilder
+}
+
+type createArguments struct {
+	BundleType string `json:"type"`
+	Nodes      []node `json:"nodes"`
```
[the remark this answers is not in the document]
br-lewis: How should this be handled by this API? Should we only specify a body if we want a remote bundle and then use `bundleType` to determine if it's `masters`, `all`, or `agents`? Would it make more sense to have two separate endpoints for remote and local in that case?

api/rest/bundle_handler.go
```diff
@@ -110,7 +131,38 @@ func (h BundleHandler) Create(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	switch strings.ToLower(args.BundleType) {
+	case bundleTypeLocal:
+		done := make(chan []string)
+		h.createLocalBundle(ctx, &bundle, done, dataFile, stateFilePath)
+	case bundleTypeRemote:
+		err = h.createRemoteBundle(ctx, &bundle, args.Nodes, dataFile, stateFilePath)
+		if err != nil {
+			writeJSONError(w, http.StatusBadRequest, err)
```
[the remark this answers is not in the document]
br-lewis: I'm not sure what you mean, if `createRemoteBundle` returns an error then the endpoint will return a 400 which you seemed to be agreeing with in your comment below. I think it makes some sense to inline it but regardless, I think the appropriate status code for all the errors it throws would be something in the 400 range.

api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+	if resp.StatusCode == http.StatusNotFound {
+		return "", fmt.Errorf("bundle %s not known to node %s", id, url)
+	} else if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("received unexpected status %d", resp.StatusCode)
+	}
+
+	// the '*' will be swapped out with a random string by ioutil.TempFile
+	destinationName := fmt.Sprintf("bundle-%s-*.zip", id)
+	// This will use the default temp directory from os.TempDir
+	destinationFile, err := ioutil.TempFile("", destinationName)
```
janisz: I think we should use configurable workdir as a temp dir or (better) accept output path as an function argument (like `curl -o`. I think we can use this client later in `dcos-cli`
```suggestion
	destinationFile, err := ioutil.TempFile(d.workdir, destinationName)
```

api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return "", nil
+	}
+	defer destinationFile.Close()
+
+	_, err = io.Copy(destinationFile, resp.Body)
+	if err != nil {
+		return "", err
```
janisz:
```suggestion
		return "", fmt.Errorf("could not save response from %s: %s", url, err)
```

api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	resp, err := d.client.Do(request)
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusNotFound {
+		return "", fmt.Errorf("bundle %s not known to node %s", id, url)
```
janisz: I think we don't need special error for this

api/rest/client.go
janisz: We should add more context to error
```suggestion
		return "", fmt.Errorf("could not save response from %s: %s", url, err)
```

api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	err = json.NewDecoder(resp.Body).Decode(&bundle)
+	if err != nil {
+		return nil, err
+	}
+
+	return &bundle, nil
+}
+
+func (d DiagnosticsClient) GetFile(ctx context.Context, node string, id string) (string, error) {
```
janisz: I think we should alias `string` to `BundleID` so our code will not be stringly typed and it will be easier to understand what type of id we are dealing with here

api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+// Client is an interface that can talk with dcos-diagnostics REST API and manipulate remote bundles
+type Client interface {
+	// Create ask node url to start bundle creation process with given id
+	Create(ctx context.Context, node string, id string) (*Bundle, error)
+	// Status returns status of bundle with given id on node at the given url
+	Status(ctx context.Context, node string, id string) (*Bundle, error)
+	// GetFile downloads bundle file of bundle with given id from node at the given
+	// url and save it to local filesystem. It returns path where file is stored or
+	// error if there were a problem.
+	GetFile(ctx context.Context, node string, id string) (path string, err error)
```
janisz: We should allow user to define where bundle should be stored
```suggestion
	GetFile(ctx context.Context, node string, id string, output string) (path string, err error)
```

janisz: We should allow user to define where bundle should be stored
```suggestion
	GetFile(ctx context.Context, node string, id string, output string) (path string, err error)
```

janisz: We should allow user to define where bundle should be stored
```suggestion
	GetFile(ctx context.Context, node string, id string, output string) (path string, err error)
```

api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+type BundleStatus struct {
+	ID   string
```
janisz: Why `ID` is exported?

api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+
+// NewDiagnosticsClient constructs a diagnostics client
+func NewDiagnosticsClient() DiagnosticsClient {
+	return DiagnosticsClient{
+		client: &http.Client{},
+	}
+}
+
+func (d DiagnosticsClient) Create(ctx context.Context, node string, id string) (*Bundle, error) {
+	url := fmt.Sprintf("%s/system/health/v1/diagnostics/%s", node, id)
```
janisz: Extract `/system/health/v1/diagnostics/` to constants

api/rest/bundle_handler.go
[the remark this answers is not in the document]
janisz: I agree this could be 400 error but it's hard to tell because you need to read underlying function to understand understand it. If we inline it it will be clear that 400 is reported because user posted no nodes.

api/rest/bundle_handler.go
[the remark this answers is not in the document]
janisz: I'm not sure how to handle it properly. I think we should stick with a single endpoint for remote and local. I think we can accept `nodes` as a list of strings. Then we can detect their roles and IPs.

Examples:
- `["leader"]` - only leader (do we need it?)
- `["masters"]` - only masters (the same works for `nodes`
- `["masters", "nodes"]` = `["all"`]` - whole cluster
- `["masters", "192.168.0.1"]` - returns only masters and one additional node (we need to check if selected node was not included in `masters` to prevent creating bundle twice
- `["masters", "192.168.0.1", "node1", "be6587c6-4838-4a4b-904a-d41c5ed9d501"]` 

@drewvanstone @makkes What do you think?

PR #125, opened 2019-06-17
Add client for DCOS_OSS-5101

br-lewis:


api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+	Status(ctx context.Context, node string, id string) (*Bundle, error)
+	// GetFile downloads bundle file of bundle with given id from node at the given
+	// url and save it to local filesystem. It returns path where file is stored or
+	// error if there were a problem.
+	GetFile(ctx context.Context, node string, id string) (path string, err error)
+}
+
+type BundleStatus struct {
+	ID   string
+	node node
```
janisz: We need node definition here

api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+// Client is an interface that can talk with dcos-diagnostics REST API and manipulate remote bundles
+type Client interface {
+	// Create ask node url to start bundle creation process with given id
+	Create(ctx context.Context, node string, id string) (*Bundle, error)
+	// Status returns status of bundle with given id on node at the given url
+	Status(ctx context.Context, node string, id string) (*Bundle, error)
+	// GetFile downloads bundle file of bundle with given id from node at the given
+	// url and save it to local filesystem. It returns path where file is stored or
+	// error if there were a problem.
+	GetFile(ctx context.Context, node string, id string) (path string, err error)
```
janisz:
```suggestion
	GetFile(ctx context.Context, node string, id string, output string) err error
```

api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+	url := fmt.Sprintf("%s/system/health/v1/diagnostics/%s", node, id)
+
+	logrus.WithField("bundle", id).WithField("url", url).Info("sending bundle creation request")
+
+	type payload struct {
+		Type string `json:"type"`
+	}
+
+	body, err := json.Marshal(payload{
+		Type: "LOCAL",
```
janisz: Move it to constant

api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+	"io"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/sirupsen/logrus"
+)
+
+// Client is an interface that can talk with dcos-diagnostics REST API and manipulate remote bundles
+type Client interface {
+	// Create ask node url to start bundle creation process with given id
```
makkes:
```suggestion
	// Create requests the given node to start a bundle creation process with that is identified by the given ID
```

api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+	Create(ctx context.Context, node string, id string) (*Bundle, error)
```
makkes: Is it possible to make `node` be of type `net/url.URL`?

makkes:
```suggestion
	Create(ctx context.Context, node string, ID string) (*Bundle, error)
```
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+	// Status returns status of bundle with given id on node at the given url
```
makkes:
```suggestion
	// Status returns the status of the bundle with the given ID on the given node
```

api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+	Status(ctx context.Context, node string, id string) (*Bundle, error)
```
makkes:
```suggestion
	Status(ctx context.Context, node string, ID string) (*Bundle, error)
```

api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+	// GetFile downloads bundle file of bundle with given id from node at the given
```
makkes:
```suggestion
	// GetFile downloads the bundle file of the bundle with the given ID from the node
```

api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+	// url and save it to local filesystem. It returns path where file is stored or
```
makkes:
```suggestion
	// and saves it to the local filesystem. It returns the path where the file is stored or
```

api/rest/client.go
```diff
@@ -0,0 +1,159 @@
...
+	// error if there were a problem.
```
makkes:
```suggestion
	// error if there was a problem.
```

api/rest/client.go
makkes:
```suggestion
	GetFile(ctx context.Context, node string, ID string) (path string, err error)
```

api/rest/client.go
makkes: I'd prefer to name this method `CreateBundle` to make it clear what this is about. It's also more in line with the `GetFile` method's name.

api/rest/client.go
```diff
@@ -62,16 +63,21 @@ func (d DiagnosticsClient) Create(ctx context.Context, node string, id string) (
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("unexpected HTTP status code received: %d", resp.StatusCode)
+		body, e := ioutil.ReadAll(resp.Body)
```
makkes: Umm, no. We're reading the full body here just to cut off everything beyond the first 100 characters. Why not read only the first 100 bytes to begin with?

api/rest/client.go
janisz: Good idea

api/rest/client.go
janisz: Thanks, fixed

api/rest/client.go
```diff
@@ -0,0 +1,158 @@
...
+
+	"github.com/sirupsen/logrus"
+)
+
+const bundlesEndpoint = "/system/health/v1/diagnostics"
+
+// Client is an interface that can talk with dcos-diagnostics REST API and manipulate remote bundles
+type Client interface {
+	// CreateBundle requests the given node to start a bundle creation process with that is identified by the given ID
+	CreateBundle(ctx context.Context, node string, ID string) (*Bundle, error)
```
janisz: I haven't spot it before. Why do we have pointer here? It's a small struct so IMO it should be a value

PR #126, opened 2019-06-18
Pull coordinator out to separate PR and make necessary changes for it to compile

br-lewis:


api/rest/coordinator_test.go
```diff
@@ -0,0 +1,143 @@
...
+		expected = append(expected,
+			BundleStatus{ID: id, node: n},
+			BundleStatus{ID: id, node: n, done: true},
+		)
+	}
+	s := c.Create(context.TODO(), "id", testNodes)
+
+	var statuses []BundleStatus
+
+	assert.Equal(t, numberOfWorkers, runtime.NumGoroutine()-numGoroutine)
```
br-lewis: I saw flaky failures due to this assert and I don't think it's going to be entirely reliable to check that the number of additional goroutines is some exact amount.

api/rest/coordinator.go
```diff
@@ -0,0 +1,227 @@
...
+type ParallelCoordinator struct {
+	client Client
+}
+
+// job is a function that will be called by worker. The output will be added to results chanel
+type job func() BundleStatus
+
+// worker is a function that will run incoming jobs from jobs channel and put jobs output to results chan
+func worker(_ context.Context, jobs <-chan job, results chan<- BundleStatus) {
+	for j := range jobs {
```
br-lewis: From what I can see, we never close the jobs channel so these goroutines will never end.

janisz: This gorutine should be closed when contex is canceled

br-lewis: Is that something that we'll have to do ourselves or will we let the 10 minute timeout on the context handle that?

api/rest/coordinator.go
```diff
@@ -0,0 +1,231 @@
...
+	"io/ioutil"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+const numberOfWorkers = 10
+
+type BundleStatus struct {
+	ID   string
```
janisz:
```suggestion
	id   string
```

api/rest/coordinator.go
```diff
@@ -0,0 +1,231 @@
...
+	node node
+	done bool
+	err  error
+}
+
+// Coordinator is an interface to coordinate the creation of diagnostics bundles
+// across a cluster of nodes
+type Coordinator interface {
+	Create(ctx context.Context, id string, nodes []node) <-chan BundleStatus
```
janisz: I think this interface should use nodes in the same way as Client so it should be just `host:port` string not a struct
```suggestion
	Create(ctx context.Context, id string, nodes []string) <-chan BundleStatus
```

api/rest/coordinator.go
```diff
@@ -0,0 +1,231 @@
...
+
+// job is a function that will be called by worker. The output will be added to results chanel
+type job func() BundleStatus
+
+// worker is a function that will run incoming jobs from jobs channel and put jobs output to results chan
+func worker(ctx context.Context, jobs <-chan job, results chan<- BundleStatus) {
+	for {
+		select {
+		case <-ctx.Done():
+			break
```
janisz: `break` will only escape from current block (`select`)
```suggestion
			return
```

api/rest/coordinator.go
```diff
@@ -0,0 +1,231 @@
...
+	return ParallelCoordinator{
+		client: client,
+	}
+}
+
+// Create starts bundle creation process. Creation process could be monitor on returned channel.
+func (c ParallelCoordinator) Create(ctx context.Context, id string, nodes []node) <-chan BundleStatus {
+
+	jobs := make(chan job)
+	statuses := make(chan BundleStatus)
```
janisz: I think this should be buffered channels. We can leave it like this for now but in the future we should revisit it

api/rest/coordinator.go
```diff
@@ -0,0 +1,231 @@
...
+		tmpNode := n
+		jobs <- func() BundleStatus {
+			//TODO(janisz): Handle context
+
+			// because this will also make a request from the coordinating master, we
+			// can't tell it to make a local bundle with the same ID this guarantees that
+			// the coordinating master's local bundle has a different ID.
+			// from this point, the full <ip>-<id> will be carried in the BundleStatus ID field so this does not
+			// need to be recalculated in the coordinator
+			fullID := fmt.Sprintf("%s-%s", tmpNode.IP, id)
```
janisz: Why not using the same ID for all local bundles?

api/rest/coordinator.go
```diff
@@ -0,0 +1,231 @@
...
+		}
+	}
+
+	return mergeZips(bundleID, bundles)
+}
+
+func mergeZips(bundleID string, bundlePaths []string) (string, error) {
+
+	bundleName := fmt.Sprintf("bundle-%s-*.zip", bundleID)
+	mergedZip, err := ioutil.TempFile("", bundleName)
```
janisz: We can't do it like this. User configure workdir and we should use it. How about creating a function interface that returns `TempFile` it could be injected into coordinator and easily mocked in tests

api/rest/coordinator.go
```diff
@@ -0,0 +1,231 @@
...
+	// If bundle is in terminal state (its state won't change)
+	if bundle.Status == Done || bundle.Status == Deleted || bundle.Status == Canceled {
+		logrus.WithField("IP", node.IP).Info("Node bundle is finished.")
+		// mark it as done
+		return BundleStatus{ID: id, node: node, done: true}
+	}
+	// If bundle is still in progress (InProgress, Unknown or Started)
+	// then schedule next check in given time
+	// It will only add check to job queue so interval might increase but it's OK.
+	time.AfterFunc(time.Second, statusCheck)
```
janisz: Make it configurable. At production it should be couple of seconds but in tests it should be millisecond.

api/rest/coordinator.go
janisz: I think waiting for timeout is fine

api/rest/coordinator.go
br-lewis: We have a few possibilities here but the main thing is to avoid using the same ID as the remote bundle, this was just the quickest to write.

api/rest/coordinator.go
br-lewis: I intentionally didn't do that pretty much just for logging purposes. Otherwise every node would be referred to as `http(s)://<ip>:<port>` which seems to be a fair amount of unnecessary information just to figure out which node the line is referring to though that also means we're passing around IP and role information that isn't functionally necessary.

I don't have strong feelings about this so I'm fine with changing it.

api/rest/coordinator.go
br-lewis: I would prefer to make the working directory part of the `ParallelCoordinator` struct. I don't think we need to introduce an interface and two implementations for this.

api/rest/coordinator.go
br-lewis: In looking at fixing the temporary files, I think we have a decent reason to keep using `node` instead of `string`. Currently bundles name the node directories `<ip>_<role>` which is probably something we'll want to keep so it's easy to see what role the node has. This would also solve the need for generating temp filenames because we now can have a deterministic and unique filename for each zip in the working directory of the bundle. So I think we should stick with `node` and do that. What do you think?

janisz: Yeah, that's a good point

api/rest/coordinator.go
janisz: OK, if we are going to stick with `node` we can have deterministic names for directories

api/rest/coordinator.go
```diff
@@ -0,0 +1,233 @@
...
+		go worker(ctx, jobs, statuses)
+	}
+
+	for _, n := range nodes {
+		logrus.WithField("IP", n.IP).Info("Sending creation request to node.")
+
+		// necessary to prevent the closure from giving the same node to all the calls
+		tmpNode := n
+		jobs <- func() BundleStatus {
+			//TODO(janisz): Handle context
```
janisz: Fixit

api/rest/coordinator.go
makkes: I still don't get why we don't use UUIDs here as I asked in https://github.com/dcos/dcos-diagnostics/pull/117#discussion_r290715769 and as the design doc says. So either we change the design doc or just use UUIDs.

api/rest/coordinator.go
```diff
@@ -0,0 +1,253 @@
...
+			fullID := fmt.Sprintf("%s-%s", tmpNode.IP, id)
+			return c.createBundle(ctx, tmpNode, fullID, jobs)
+		}
+	}
+
+	return statuses
+}
+
+func (c ParallelCoordinator) Collect(ctx context.Context, bundleID string, numBundles int, statuses <-chan BundleStatus) (string, error) {
+
```
makkes:
```suggestion
bundles will contain the local path to all of the collected bundle files after waiting for all bundles to be ready (excluding those that failed to be created).
```

api/rest/bundle_handler.go
```diff
@@ -72,6 +73,12 @@ type BundleHandler struct {
 	bundleCreationTimeout time.Duration         // limits how long bundle creation could take
 }
 
+type node struct {
```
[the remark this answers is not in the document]
br-lewis: It was originally here since it's part of the input arguments for the bundlehandler creation endpoint so I decided to put it here now rather than put it in coordinator then move it in the next PR.

api/rest/coordinator_test.go
```diff
@@ -25,6 +25,8 @@ func TestCoordinator_CreatorShouldCreateAbundleAndReturnUpdateChan(t *testing.T)
 
 	ctx := context.TODO()
 
+	localBundleID := uuid.New()
```
makkes: Nitpicking: For the tests we wouldn't even need a UUID except we actually checked the format and rejected non-UUIDv4 IDs.

api/rest/coordinator.go
```diff
@@ -0,0 +1,256 @@
...
+		case <-ctx.Done():
+			return "", fmt.Errorf("context canceled before all node bundles finished")
+		case s := <-statuses:
+			if !s.done {
+				logrus.WithError(s.err).WithField("IP", s.node.IP).WithField("ID", s.id).Info("Got status update. Bundle not ready.")
+				continue
+			}
+			if s.err != nil {
+				logrus.WithError(s.err).WithField("IP", s.node.IP).WithField("ID", s.id).Warn("Bundle errored")
+				// TODO(br-lewis): this should probably be noted in the generated bundle and not just printed in the journal
```
makkes: What about this?

br-lewis: This would be a very large change since currently the coordinator doesn't add any output in the resulting bundle and should probably just be done separately.

makkes:
```suggestion
				// TODO (https://jira.mesosphere.com/browse/DCOS_OSS-5303): this should be noted in the generated bundle and not just printed in the journal
```

api/rest/coordinator.go
```diff
@@ -0,0 +1,256 @@
...
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+const numberOfWorkers = 10
+const contextDoneErrMsg = "bundle creation context finished before bundle creation finished"
+
+// BundleStatus tracks the status of local bundle creation requests
+type BundleStatus struct {
```
janisz: I think this could be package private 
```suggestion
type bundleStatus struct {
```

api/rest/coordinator.go
```diff
@@ -0,0 +1,256 @@
...
+	id   string
+	node node
+	done bool
+	err  error
+}
+
+// Coordinator is an interface to coordinate the creation of diagnostics bundles
+// across a cluster of nodes
+type Coordinator interface {
```
janisz: I think this could be package private or we should move Coordinator to separate package
```suggestion
type coordinator interface {
```

api/rest/coordinator.go
```diff
@@ -0,0 +1,256 @@
...
+	// on the returned channel.
+	CreateBundle(ctx context.Context, id string, nodes []node) <-chan BundleStatus
+	// CollectBundle waits until all the nodes' bundles have finished, downloads,
+	// and merges them. The resulting bundle zip file path is returned.
+	CollectBundle(ctx context.Context, bundleID string, numBundles int, statuses <-chan BundleStatus) (string, error)
+}
+
+// ParallelCoordinator implements Coordinator interface to coordinate bundle
+// creation across a cluster, parallelized across multiple goroutines
+type ParallelCoordinator struct {
```
janisz: ditto

api/rest/coordinator.go
```diff
@@ -0,0 +1,256 @@
...
+		case <-ctx.Done():
+			return
+		case j := <-jobs:
+			results <- j()
+		}
+	}
+}
+
+// NewParallelCoordinator constructs a new parallelcoordinator.
+func NewParallelCoordinator(client Client, interval time.Duration, workDir string) ParallelCoordinator {
```
janisz: Please add a comment what `interval` means in this context

janisz: Or add this comment to `ParallelCoordinator` struct definition

api/rest/coordinator.go
```diff
@@ -0,0 +1,256 @@
...
+	"io"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+const numberOfWorkers = 10
+const contextDoneErrMsg = "bundle creation context finished before bundle creation finished"
```
janisz: This const is useed only with `fmt.Errorf` please change it to `error`

api/rest/coordinator.go
```diff
@@ -0,0 +1,256 @@
...
+func (c ParallelCoordinator) CollectBundle(ctx context.Context, bundleID string, numBundles int, statuses <-chan BundleStatus) (string, error) {
+
+	// holds the paths to the downloaded local bundles before merging
+	var bundlePaths []string
+
+	finishedBundles := 0
+	for finishedBundles < numBundles {
+		select {
+		case <-ctx.Done():
+			return "", fmt.Errorf("context canceled before all node bundles finished")
```
janisz: I'm not sure if this is a valid approach. What if we have problems with a single node that is not responding? Then a whole bundle will not be generated. I think we should return generated bundle with information that timeout occured

api/rest/coordinator.go
```diff
@@ -0,0 +1,256 @@
...
+		case s := <-statuses:
+			if !s.done {
+				logrus.WithError(s.err).WithField("IP", s.node.IP).WithField("ID", s.id).Info("Got status update. Bundle not ready.")
+				continue
+			}
+			if s.err != nil {
+				logrus.WithError(s.err).WithField("IP", s.node.IP).WithField("ID", s.id).Warn("Bundle errored")
+				// TODO (https://jira.mesosphere.com/browse/DCOS_OSS-5303): this should be noted in the generated bundle and not just printed in the journal
```
janisz: We can generate a mock bundle with a single error file

janisz: Or better: collect all errors and log them in summarry error log.

api/rest/coordinator.go
```diff
@@ -0,0 +1,256 @@
...
+	// Schedule bundle status check
+	jobs <- func() BundleStatus {
+		return c.waitForDone(ctx, node, id, c.statusCheckInterval, jobs)
+	}
+
+	// Return undone status with no error.
+	return BundleStatus{id: id, node: node}
+}
+
+func (c ParallelCoordinator) waitForDone(ctx context.Context, node node, id string, interval time.Duration, jobs chan<- job) BundleStatus {
```
janisz: `ParallelCoordinator` already has status check interval we don't need it as an argument
```suggestion
func (c ParallelCoordinator) waitForDone(ctx context.Context, node node, id string, jobs chan<- job) BundleStatus {
```

api/rest/coordinator.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	logrus.WithField("IP", node.IP).Info("Checking bundle status on node.")
+	// Check bundle status
+	bundle, err := c.client.Status(ctx, node.baseURL, id)
+	// If error
+	if err != nil {
+		// then schedule next check in given time.
+		// It will only add check to job queue so interval might increase but it's OK.
+		time.AfterFunc(interval, statusCheck)
```
janisz:
```suggestion
		time.AfterFunc( c.statusCheckInterval, statusCheck)
```

api/rest/coordinator_test.go
janisz: Are we using github.com/google/uuid only for tests? I think we don't need that dependency at this step.

api/rest/coordinator.go
janisz: We need a test case for this case

api/rest/coordinator.go
```diff
@@ -0,0 +1,256 @@
...
+				logrus.WithError(s.err).WithField("IP", s.node.IP).WithField("ID", s.id).Warn("Bundle errored")
+				// TODO (https://jira.mesosphere.com/browse/DCOS_OSS-5303): this should be noted in the generated bundle and not just printed in the journal
+				finishedBundles++
+				continue
+			}
+
+			bundlePath := filepath.Join(c.workDir, nodeBundleFilename(s.node))
+			err := c.client.GetFile(ctx, s.node.baseURL, s.id, bundlePath)
+			if err != nil {
+				logrus.WithError(err).WithField("IP", s.node.IP).WithField("ID", s.id).Warn("Could not download file")
```
janisz: I think we should `continue` on this step. If there is an error we may not be able to merge zip if it does not exist. We need a test case for this case
```suggestion
				logrus.WithError(err).WithField("IP", s.node.IP).WithField("ID", s.id).Warn("Could not download file")
				finishedBundles++
				continue
```

api/rest/coordinator.go
```diff
@@ -0,0 +1,256 @@
...
+		if err != nil {
+			return "", err
+		}
+	}
+
+	return mergedZip.Name(), nil
+}
+
+func appendToZip(writer *zip.Writer, path string) error {
+	r, err := zip.OpenReader(path)
```
janisz: please add a test with malformed zip

api/rest/coordinator.go
```diff
@@ -0,0 +1,256 @@
...
+		}
+		rc.Close()
+	}
+
+	return nil
+}
+
+func (c ParallelCoordinator) createBundle(ctx context.Context, node node, id string, jobs chan<- job) BundleStatus {
+	_, err := c.client.CreateBundle(ctx, node.baseURL, id)
+	if err != nil {
```
janisz: Please add a test with not responding node

api/rest/coordinator.go
```diff
@@ -0,0 +1,256 @@
...
+		jobs <- func() BundleStatus {
+			return c.waitForDone(ctx, node, id, c.statusCheckInterval, jobs)
+		}
+	}
+
+	logrus.WithField("IP", node.IP).Info("Checking bundle status on node.")
+	// Check bundle status
+	bundle, err := c.client.Status(ctx, node.baseURL, id)
+	// If error
+	if err != nil {
```
janisz: We need a test for this

api/rest/coordinator.go
```diff
@@ -0,0 +1,256 @@
...
+		logrus.WithField("IP", node.IP).Info("Node bundle is finished.")
+		// mark it as done
+		return BundleStatus{id: id, node: node, done: true}
+	}
+	// If bundle is still in progress (InProgress, Unknown or Started)
+	// then schedule next check in given time
+	// It will only add check to job queue so interval might increase but it's OK.
+	time.AfterFunc(c.statusCheckInterval, statusCheck)
+	// Return undone status with no error. Do not mark bundle as done yet. It might change it status
+	return BundleStatus{id: id, node: node}
```
janisz: It looks like we never get to this point in tests. We should check it since it's a part of happy path

janisz:
@drewvanstone Can you look at this?

api/rest/coordinator.go
br-lewis: Since we don't have any sort of remote bundle logging at the moment, I thought this would be too big of a task given that this PR is already quite large so put it in as another task to do separate from this.

makkes: Yes, let's use DCOS_OSS-5303 for that, pls.

api/rest/coordinator.go
br-lewis: Interestingly, making this private causes our linter to fail. Since we only test `parallelCoordinator` as itself and not as a `coordinator` I think Go isn't bothering to consider that it implements the `coordinator` interface so it assumes it isn't used.

api/rest/coordinator.go
br-lewis: I think this is best handled as part of https://jira.mesosphere.com/browse/DCOS_OSS-5303. In the meantime, I'm going to have it merge the zips that have already been downloaded and log that the context was cancelled.

api/rest/coordinator.go
```diff
@@ -24,9 +24,12 @@ type bundleStatus struct {
 	err  error
 }
 
-// Coordinator is an interface to coordinate the creation of diagnostics bundles
+// golangcli-lint marks this as dead code because nothing uses the interface
+// yet. this will be removed with the bundle handler PR
+//nolint:deadcode
```
janisz: Remember to remove it after using coordinator in handler

api/rest/coordinator.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+	defer mergedZip.Close()
+
+	zipWriter := zip.NewWriter(mergedZip)
+	defer zipWriter.Close()
+
+	for _, p := range bundlePaths {
+		err = appendToZip(zipWriter, p)
+		if err != nil {
+			return "", err
```
janisz: Do we wan't to stop bundle creation if one bundle file is invalid? I think we should collect that error and add it to error summary

api/rest/coordinator_test.go
```diff
@@ -0,0 +1,293 @@
...
+	results := []bundleStatus{}
+
+	for i := 0; i < len(expected); i++ {
+		results = append(results, <-statuses)
+	}
+
+	for _, s := range results {
+		assert.Contains(t, expected, s)
+	}
+}
```
janisz: We need one test for context cancel

api/rest/coordinator.go
br-lewis: You're probably right that we don't. I'll add a comment with that other ticket about the summary report.

api/rest/coordinator.go
```diff
@@ -0,0 +1,271 @@
...
+	// be checked
+	statusCheckInterval time.Duration
+	workDir             string
+}
+
+// job is a function that will be called by the worker function. The output will be added to results channel
+type job func() bundleStatus
+
+// worker is a function that will run incoming jobs from jobs channel and put jobs output to results chan
+func worker(ctx context.Context, jobs <-chan job, results chan<- bundleStatus) {
```
janisz:
```suggestion
func worker(jobs <-chan job, results chan<- bundleStatus) {
```

api/rest/coordinator.go
```diff
@@ -0,0 +1,271 @@
...
+	for {
+		select {
+		case j := <-jobs:
```
janisz:
```suggestion
		for j :=  range jobs {
```

And we need to find a way to close jobs channel

janisz: How about having information about number of expected number of bundles and count finished bundles and exit when no more job need to be done?

api/rest/coordinator.go
janisz: Add a TODO with ticket number

api/rest/coordinator.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+// job is a function that will be called by the worker function. The output will be added to results channel
+type job func(context.Context) bundleStatus
+
+// worker is a function that will run incoming jobs from jobs channel and put jobs output to results chan
+func worker(ctx context.Context, jobs <-chan job, results chan<- bundleStatus, quit <-chan struct{}) {
+	for {
+		select {
+		case <-quit:
```
janisz: In previous commit you remove context, now it's back. Why just not rely on contex timeout instead of adding new channel?

br-lewis: I think I was thinking about how this would work incorrectly, doing a context done check there should be fine since the waiting `CollectBundle` loop would still run and create what it can of the bundle.

br-lewis: Ok never mind, that causes a deadlock. I think because exiting the worker function will leave behind goroutines that try to push to the `jobs` channel but nothing is reading from it. This is showing up pretty consistently in the tests because jobs isn't buffered right now.

api/rest/coordinator.go
```diff
@@ -0,0 +1,284 @@
...
+			// even if the bundle finished with an error, it's now finished so increment finishedBundles
+			finishedBundles++
+			if s.err != nil {
+				// TODO (https://jira.mesosphere.com/browse/DCOS_OSS-5303): this should be noted in the generated bundle and not just printed in the journal
+				logrus.WithError(s.err).WithField("IP", s.node.IP).WithField("ID", s.id).Warn("Bundle errored")
+				continue
+			}
+
+			bundlePath := filepath.Join(c.workDir, nodeBundleFilename(s.node))
+			err := c.client.GetFile(ctx, s.node.baseURL, s.id, bundlePath)
```
janisz: We should add retry here

PR #127, opened 2019-06-19
Add team individuals as a codeowners

janisz:


CODEOWNERS
```diff
@@ -1 +1,5 @@
 * @dcos/cluster-ops-fnd
+* @makkes
+* @drewanstone
+* @janisz
+* @br-levis
```
br-lewis:
```suggestion
* @br-lewis
```

janisz: Thanks

CODEOWNERS
```diff
@@ -1 +1,5 @@
 * @dcos/cluster-ops-fnd
```
makkes: can this not be removed?

janisz: We have a check that require person from team to approve

PR #128, opened 2019-06-28
Add changes in Bundle Handler for remote bundle creation

br-lewis:


api/rest/bundle_handler.go
```diff
@@ -31,6 +32,9 @@ const (
 
 	filePerm = 0600
 	dirPerm  = 0700
+
+	bundleTypeLocal  = "local"
+	bundleTypeRemote = "remote"
```
janisz: I thought we just agreed we don't need type

br-lewis:
This is being done in multiple PRs

PR #131, opened 2019-07-12
Add cluster level API for diagnostics bundles

br-lewis:
Addresses https://jira.mesosphere.com/browse/DCOS_OSS-5314 by adding a cluster bundle handler that manages bundles on master nodes and can create a cluster bundle.

`ClusterBundleHandler` is pretty much just a wrapper around the node level APIs of all the masters, except when asked to create a bundle as then it collect bundles from all cluster nodes.

PR #132, opened 2019-07-17
Do not merge yet - Adds debug information to cluster bundles

br-lewis:
This adds a JSON file with information about whether the node bundles succeeded and has the error message if they didn't.

Related ticket: DCOS_OSS-5303 (this'll be a link when Jira comes back up).

This builds off of https://github.com/dcos/dcos-diagnostics/pull/131 so shouldn't be merged until that's merged.

### DIFF
--- a/api/rest/coordinator_test.go
+++ b/api/rest/coordinator_test.go
@@ -49,7 +49,9 @@ func TestCoordinator_CreatorShouldCreateAbundleAndReturnUpdateChan(t *testing.T)
 			BundleStatus{id: localBundleID, node: n, done: true},
 		)
 	}
-	s := c.CreateBundle(context.TODO(), localBundleID, testNodes)
+
+	report := newBundleReport("cluster-bundle")
+	s := c.CreateBundle(context.TODO(), localBundleID, testNodes, report)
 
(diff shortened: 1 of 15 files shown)


api/rest/coordinator_test.go
```diff
@@ -110,9 +112,10 @@ func TestCoordinatorCreateAndCollect(t *testing.T) {
 		client.On("GetFile", ctx, testData.n.baseURL, localBundleID, testData.zipPath).Return(nil)
 	}
 
-	statuses := c.CreateBundle(ctx, localBundleID, []node{node1, node2, node3})
+	report := newBundleReport("cluster-bundle")
+	statuses := c.CreateBundle(ctx, localBundleID, []node{node1, node2, node3}, report)
```
janisz: I do not like this approach. Reports should be returned by Coordinator functions and merged at the end in to a single report. With approach like this we have a shared state object in highly parallel world. This may not be easy to handle.

janisz: @makkes @drewvanstone What do you think?

janisz:
Closing in favor of https://github.com/dcos/dcos-diagnostics/pull/134

PR #133, opened 2019-07-19
Add `mesos-state` command to CLI

janisz:
Fixes: https://jira.mesosphere.com/browse/DCOS-56380

### DIFF
--- a/cmd/daemon.go
+++ b/cmd/daemon.go
@@ -24,9 +24,9 @@ import (
 
 	"github.com/dcos/dcos-diagnostics/api"
 	"github.com/dcos/dcos-diagnostics/api/rest"
+	diagDcos "github.com/dcos/dcos-diagnostics/dcos"
 	"github.com/dcos/dcos-diagnostics/util"
 
-	diagDcos "github.com/dcos/dcos-diagnostics/dcos"
 	"github.com/dcos/dcos-go/dcos"
 	"github.com/dcos/dcos-go/dcos/http/transport"
 	"github.com/dcos/dcos-go/dcos/nodeutil"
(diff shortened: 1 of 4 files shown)


cmd/daemon.go
```diff
@@ -53,41 +53,18 @@ var daemonCmd = &cobra.Command{
...
-	var transportOptions []transport.OptionTransportFunc
-	if defaultConfig.FlagCACertFile != "" {
-		transportOptions = append(transportOptions, transport.OptionCaCertificatePath(defaultConfig.FlagCACertFile))
-	}
-	if defaultConfig.FlagIAMConfig != "" {
-		transportOptions = append(transportOptions, transport.OptionIAMConfigPath(defaultConfig.FlagIAMConfig))
-	}
-
-	tr, err := transport.NewTransport(transportOptions...)
+	tr, err := initTransport()
```
drewvanstone: 👍

cmd/mesos_state.go
```diff
@@ -0,0 +1,50 @@
...
+	"net/url"
+	"strconv"
+
+	"github.com/dcos/dcos-diagnostics/util"
+	"github.com/dcos/dcos-go/dcos"
+	"github.com/spf13/cobra"
+)
+
+// daemonCmd represents the daemon command
+var stateCmd = &cobra.Command{
```
makkes: Please add a test for this. IMHO it's enough to have one tests that just runs the cmd using a mock HTTP server and checks whether the body is just returned. To do this I would seperate out the logic from the `RunE` function into a function of its own and run the tests against that then.

janisz: Can you take a look now?

cmd/mesos_state_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_getMesosState(t *testing.T) {
+	tr := roundTripFunc(func(r *http.Request) (*http.Response, error) {
+		assert.Equal(t, r.URL.String(), "http://leader.mesos:5050/state")
+		return &http.Response{
+			StatusCode: http.StatusOK,
+			Body:       ioutil.NopCloser(strings.NewReader(`{}`)),
```
makkes: I would return sth. more unique such as `{"this": "and that"}` to be sure the `getMesosState` function doesn't swallow the body and just always returns an empty JSON string.

cmd/mesos_state.go
makkes: I would actually love to see a test that tests this command itself so that we can be sure that this command prints the expected result. In https://github.com/makkes/gitlab-cli/blob/master/cmd/issue/inspect/inspect.go#L15 I even passed an `io.Writer` to the function so that the cmd itself is only one line of code and the rest is completely testable.

I find this important esp. since we don't have e2e/integration tests for this.

PR #137, opened 2019-08-01
Allow configuration of `detect_ip` script

janisz:
Add `--ip-discovery-command` config options that accepts path to `detect_ip` script.

Fixes: https://jira.mesosphere.com/browse/DCOS_OSS-5384


janisz:
There are no test in this package but I can add them

janisz:
@makkes I added test

PR #144, opened 2019-08-21
Fix race on writing to bundle state file

janisz:
* Add mutex to synchronize all writes to a file. The mutext is shared for all writes but since API need to support small amount of request it's fine.
* Do not update state file when get state detect file size differs from value stored in a file
* Sleep in E2E test to ensure everything is updated and we do not read previous version of a state file

Fixes: https://jira.mesosphere.com/browse/DCOS_OSS-5331

### DIFF
--- a/api/rest/bundle_handler.go
+++ b/api/rest/bundle_handler.go
@@ -58,7 +58,7 @@ func (realClock) Now() time.Time { return time.Now() }
 
 func NewBundleHandler(workDir string, collectors []collector.Collector, timeout time.Duration) BundleHandler {
 	return BundleHandler{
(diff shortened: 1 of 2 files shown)


api/rest/bundle_handler.go
```diff
@@ -57,6 +58,7 @@ func (realClock) Now() time.Time { return time.Now() }
 
 func NewBundleHandler(workDir string, collectors []collector.Collector, timeout time.Duration) BundleHandler {
 	return BundleHandler{
+		mutex:                 &sync.RWMutex{},
```
makkes:
```suggestion
		stateFileLock:                 &sync.RWMutex{},
```

PR #145, opened 2019-08-21
Add workdir initialization to BundleHandler and ClusterBundleHandler

br-lewis:
Addresses https://jira.mesosphere.com/browse/DCOS_OSS-5454 by having `NewBundleHandler` and `NewClusterBundleHandler` initialize `workdir` if it doesn't exist.

### DIFF
--- a/api/rest/bundle_handler.go
+++ b/api/rest/bundle_handler.go
@@ -55,13 +55,33 @@ type realClock struct{}
 
 func (realClock) Now() time.Time { return time.Now() }
 
-func NewBundleHandler(workDir string, collectors []collector.Collector, timeout time.Duration) BundleHandler {
-	return BundleHandler{
+func NewBundleHandler(workDir string, collectors []collector.Collector, timeout time.Duration) (*BundleHandler, error) {
+	f, err := os.Stat(workDir)
+	if err != nil {
+		fmt.Println("error stating workdir")
+		if os.IsNotExist(err) {
+			fmt.Println("workdir does not exist")
+			err = os.Mkdir(workDir, dirPerm)
+			if err != nil {
+				fmt.Println("unable to create")
+				logrus.WithError(err).Errorf("workDir does not exist and could not be created %s", workDir)
+				return nil, err
+			}
+		} else {
+			logrus.Errorf("Could not open workDir %s", workDir)
+			return nil, err
+		}
+	} else if f.Mode().IsRegular() {
+		logrus.Errorf("workDir exists but is not a directory: %s", workDir)
+		return nil, err
+	}
+
+	return &BundleHandler{
 		clock:                 realClock{},
 		workDir:               workDir,
 		collectors:            collectors,
 		bundleCreationTimeout: timeout,
-	}
+	}, nil
 }
 
 // BundleHandler is a struct that collects all functions
--- a/api/rest/bundle_handler.go
+++ b/api/rest/bundle_handler.go
@@ -57,20 +57,8 @@ type realClock struct{}
 func (realClock) Now() time.Time { return time.Now() }
 
 func NewBundleHandler(workDir string, collectors []collector.Collector, timeout time.Duration) (*BundleHandler, error) {
-	f, err := os.Stat(workDir)
+	err := initializeWorkDir(workDir)
 	if err != nil {
-		if os.IsNotExist(err) {
-			err = os.Mkdir(workDir, dirPerm)
-			if err != nil {
-				logrus.WithError(err).Errorf("workDir does not exist and could not be created %s", workDir)
-				return nil, err
-			}
-		} else {
-			logrus.Errorf("Could not open workDir %s", workDir)
-			return nil, err
-		}
-	} else if f.Mode().IsRegular() {
-		logrus.Errorf("workDir exists but is not a directory: %s", workDir)
 		return nil, err
 	}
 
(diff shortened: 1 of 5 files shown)


api/rest/bundle_handler.go
```diff
@@ -56,14 +56,31 @@ type realClock struct{}
 
 func (realClock) Now() time.Time { return time.Now() }
 
-func NewBundleHandler(workDir string, collectors []collector.Collector, timeout time.Duration) BundleHandler {
-	return BundleHandler{
+func NewBundleHandler(workDir string, collectors []collector.Collector, timeout time.Duration) (*BundleHandler, error) {
+	f, err := os.Stat(workDir)
+	if err != nil {
```
drewvanstone: Not sure what you guys think, but in Go, I tend to think of `else` and `else if` statements as mild code smells, and usually can find a way to work around it. In this particular case, and the case below on line 72, the else statement doesn't look to be needed at all. I personally find it a bit more readable, since you don't need to hold quite as much branching logic in mind.

```suggestion
	if err != nil {
		if os.IsNotExist(err) {
			err = os.Mkdir(workDir, dirPerm)
			if err != nil {
				logrus.WithError(err).Errorf("workDir does not exist and could not be created %s", workDir)
				return nil, err
			}
		} 
		logrus.Errorf("Could not open workDir %s", workDir)
		return nil, err
	}
		
	if f.Mode().IsRegular() {
		logrus.Errorf("workDir exists but is not a directory: %s", workDir)
		return nil, err
	}
```

api/rest/bundle_handler.go
```diff
@@ -56,14 +56,31 @@ type realClock struct{}
...
+				return nil, err
+			}
+		} else {
+			logrus.Errorf("Could not open workDir %s", workDir)
+			return nil, err
+		}
+	} else if f.Mode().IsRegular() {
+		logrus.Errorf("workDir exists but is not a directory: %s", workDir)
+		return nil, err
+	}
```
janisz: How about replacing the code above with


```go
err := os.MkdirAll(workDir, dirPerm)
if err != nil {
	return nil, fmt.Errorf("could not create %s: %s", workDir, err)
}
```

api/rest/cluster_bundle_handler_test.go
```diff
@@ -871,6 +871,24 @@ func TestRemoteBundleCreation(t *testing.T) {
...
+	coord := mockCoordinator{}
+	client := &MockClient{}
+	tools := &MockedTools{}
+	urlBuilder := MockURLBuilder{}
+	_, err = NewClusterBundleHandler(coord, client, tools, workdir, time.Millisecond, urlBuilder)
+	require.NoError(t, err)
+
+	assert.DirExists(t, workdir)
+}
+
```
janisz: I one more test to check what happens when file under given workdir path exists
```suggestion

func TestBundleDirCreationFailsWhenFileWithTheSameNameExists(t *testing.T) {
	t.Parallel()

	workdir, err := ioutil.TempFile("", "work-dir")
	require.NoError(t, err)
	_, err = NewBundleHandler(workdir.Name(), nil, time.Millisecond)
	assert.Error(t, err)
}
```

api/rest/bundle_handler.go
```diff
@@ -56,14 +56,31 @@ type realClock struct{}
...
-func NewBundleHandler(workDir string, collectors []collector.Collector, timeout time.Duration) BundleHandler {
-	return BundleHandler{
+func NewBundleHandler(workDir string, collectors []collector.Collector, timeout time.Duration) (*BundleHandler, error) {
+	f, err := os.Stat(workDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			err = os.Mkdir(workDir, dirPerm)
+			if err != nil {
+				logrus.WithError(err).Errorf("workDir does not exist and could not be created %s", workDir)
+				return nil, err
```
janisz: Logs should be handled only once. Never log and pass an error.

api/rest/bundle_handler.go
makkes: +1 https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

janisz: I concur. I also like to avoid multiple indentation. Here if we invert `os.IsNotExist` we can have even simpler code
```go
	if err != nil {
		if !os.IsNotExist(err) {
			return nil, err
		}
		
		err = os.Mkdir(workDir, dirPerm)
		if err != nil {
			return nil, err
		}
	}

	if f.Mode().IsRegular() {
		return nil, err
	}
```

api/rest/bundle_handler.go
drewvanstone: > MkdirAll creates a directory named path, along with any necessary parents, and returns nil, or else returns an error. The permission bits perm (before umask) are used for all directories that MkdirAll creates. If path is already a directory, MkdirAll does nothing and returns nil.

Nice, ya that seems to be a good idea @janisz

api/rest/cluster_bundle_handler.go
```diff
@@ -31,7 +31,26 @@ type ClusterBundleHandler struct {
 }
 
 func NewClusterBundleHandler(c Coordinator, client Client, tools dcos.Tooler, workDir string, timeout time.Duration,
-	urlBuilder dcos.NodeURLBuilder) *ClusterBundleHandler {
+	urlBuilder dcos.NodeURLBuilder) (*ClusterBundleHandler, error) {
+	f, err := os.Stat(workDir)
```
makkes: Since this is the exact same code as above in `NewBundleHandler` you should outsource this into a func, e.g. `ensureWorkDirExists()`.

PR #147, opened 2019-08-22
Delete local bundles after fetching them

janisz:
Fixes: https://jira.mesosphere.com/browse/DCOS_OSS-5470

### DIFF
--- a/api/rest/coordinator.go
+++ b/api/rest/coordinator.go
@@ -71,6 +71,11 @@ type nodeBundleReport struct {
 	Err    string `json:"error,omitempty"`
 }
 
+type bundleToDelete struct {
+	baseURL       string
+	localBundleID string
+}
+
 // job is a function that will be called by the worker function. The output will be added to results channel
 type job func(context.Context) BundleStatus
 
@@ -128,6 +133,8 @@ func (c ParallelCoordinator) CollectBundle(ctx context.Context, bundleID string,
 		Nodes: make(map[string]nodeBundleReport, numBundles),
 	}
 
+	var bundlesToDelete = make([]bundleToDelete, numBundles)
+
 	for finishedBundles := 0; finishedBundles < numBundles; {
 
 		s := <-statuses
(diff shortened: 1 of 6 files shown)


api/rest/coordinator.go
```diff
@@ -158,6 +166,18 @@ func (c ParallelCoordinator) CollectBundle(ctx context.Context, bundleID string,
 		bundlePaths = append(bundlePaths, bundlePath)
 	}
 
+	// Run cleanup in separated goroutine so it will not block bundle generation process
+	go func() {
+		for _, b := range bundlesToDelete {
```
makkes: Can we be sure that `bundlesToDelete == numBundles` at this point? Otherwise you'd have to check whether `b` is the zero value here.

janisz: Yes we can, there is a simple for loop https://github.com/dcos/dcos-diagnostics/pull/147/files/2964a8cf0fd0675cfb6b69c3ed0747da7423b5a9#diff-7b54bfd9ba87b3e1cbcc388b026985d7R138 with no `break`

api/rest/new_mock_client_test.go
```diff
@@ -4,10 +4,10 @@ import "context"
...
 type MockClient struct {
 	createBundle func(ctx context.Context, node string, ID string) (*Bundle, error)
-	status func(ctx context.Context, node string, ID string) (*Bundle, error)
-	getFile func(ctx context.Context, node string, ID string, path string) (err error)
-	list func(ctx context.Context, node string) ([]*Bundle, error)
-	delete func(ctx context.Context, node string, id string) error
+	status       func(ctx context.Context, node string, ID string) (*Bundle, error)
+	getFile      func(ctx context.Context, node string, ID string, path string) (err error)
+	list         func(ctx context.Context, node string) ([]*Bundle, error)
+	delete       func(ctx context.Context, node string, id string) error
```
drewvanstone: Think we should use `ID` here to be consistent

janisz: Fixed

PR #149, opened 2019-08-23
Check if bundle is done before it could be fetch

janisz:
Fixes: DCOS_OSS-5483

### DIFF
--- a/api/rest/bundle_handler.go
+++ b/api/rest/bundle_handler.go
@@ -241,6 +241,12 @@ func (h BundleHandler) GetFile(w http.ResponseWriter, r *http.Request) {
 		writeJSONError(w, http.StatusInternalServerError, err)
 		return
 	}
+	if bundle.Status == Deleted || bundle.Status == Canceled || bundle.Status == Failed {
+		writeJSONError(w, http.StatusNotFound,
+			fmt.Errorf("bundle %s was %s", bundle.ID, bundle.Status))
+		return
+	}
+
 	if bundle.Status != Done {
 		writeJSONError(w, http.StatusNotFound,
 			fmt.Errorf("bundle %s is not done yet (status %s), try again later", bundle.ID, bundle.Status))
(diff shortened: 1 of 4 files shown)


api/rest/cluster_bundle_handler.go
```diff
@@ -281,19 +281,21 @@ func (c *ClusterBundleHandler) Download(w http.ResponseWriter, r *http.Request)
 	var masterWithBundle node
 	found := false
 	for _, n := range masters {
-		_, err = c.client.Status(ctx, n.baseURL, id)
-		if err != nil {
-			switch err.(type) {
+		bundle, e := c.client.Status(ctx, n.baseURL, id)
```
makkes: Is there a specific reason to rename the `err` variable? I'd say it's convenience across our projects to call it `err`.

api/rest/bundle_handler.go
```diff
@@ -235,6 +235,18 @@ func (h BundleHandler) Get(w http.ResponseWriter, r *http.Request) {
...
 	vars := mux.Vars(r)
 	id := vars["id"]
+
+	bundle, err := h.getBundleState(id)
+	if err != nil {
+		writeJSONError(w, http.StatusInternalServerError, err)
+		return
+	}
+	if bundle.Status != Done {
+		writeJSONError(w, http.StatusNotFound,
```
br-lewis: If we end up with a bundle in the `Failed` state, should that change our error? Is it even possible to get that with cluster bundles?

janisz: Good point! If bundle is Canceled, Failed or Deleted we should return 404

janisz: Failed state is here https://github.com/dcos/dcos-diagnostics/pull/150

api/rest/cluster_bundle_handler.go
janisz:
```api/rest/cluster_bundle_handler.go:284:11: shadow: declaration of "err" shadows declaration at line 273 (govet)
		bundle, err := c.client.Status(ctx, n.baseURL, id)
```

janisz: I think it's better to change name than disable linter

api/rest/bundle_handler.go
janisz: Fixed

api/rest/cluster_bundle_handler.go
makkes: Then I'd prefer sth. more meaningful such as `statusErr`.

janisz: Fixed

PR #154, opened 2019-08-30
Return 200 when deleting already deleted bundle

janisz:
Fixes: https://jira.mesosphere.com/browse/DCOS-58105

### DIFF
--- a/api/rest/bundle_handler.go
+++ b/api/rest/bundle_handler.go
@@ -261,7 +261,7 @@ func (h BundleHandler) GetFile(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if bundle.Status == Deleted || bundle.Status == Canceled || bundle.Status == Failed {
-		writeJSONError(w, http.StatusNotFound,
+		writeJSONError(w, http.StatusGone,
 			fmt.Errorf("bundle %s was %s", bundle.ID, bundle.Status))
 		return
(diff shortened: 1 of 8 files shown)


janisz:
What endpoint? Get alwyas returns bundle status even when it's `DELETED`. When bundle is deleted you can't get it's file, it returns 404 https://github.com/dcos/dcos-diagnostics/blob/df3a67a0658d103d55bfd734d86ae818aeeb2e30/api/rest/bundle_handler_test.go#L722-L723

janisz:
@bamarni I think I know what you mean (the ticket has a typo 304 instead of 404). I fixed the code so now when bundle was deleted/failed/canceled we returns 410 instead of 404.

bamarni:
@janisz : It's not a typo, I was referring to https://github.com/dcos/dcos-core-cli/blob/90e03f6af36aac507e73657abf59542ad3f745b7/pkg/diagnostics/v2/client.go#L139, the client seems to expect a 304 here. But I didn't check if the API can actually return a 304, maybe it's a mistake in the client?

janisz:
Is now ok or should I change anything? There are two separate endpoints one to get bundle status and second to download file

bamarni:
@janisz : should be good, browsing through the code it seems like the line I've linked is an inconsistency in the client only.

PR #156, opened 2019-09-04
Use go 1.12 in Travis

janisz:
Go 1.13 leads to failing CI we need to investigate and fix it.

Sww: https://travis-ci.org/dcos/dcos-diagnostics/jobs/580355340


janisz:
It's here https://jira.mesosphere.com/browse/DCOS_OSS-5517

PR #159, opened 2019-10-04
Allow limiting bundle to agents or masters only

janisz:
Add option to limit diagnostic bundle to masters or agents only. 
Replace apiary API documentation with OpenApi 

Fixes:
* https://jira.mesosphere.com/browse/DCOS_OSS-5472
* https://jira.mesosphere.com/browse/DCOS_OSS-5474

### DIFF
--- a/api/rest/cluster_bundle_handler.go
+++ b/api/rest/cluster_bundle_handler.go
@@ -2,6 +2,7 @@ package rest
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -54,13 +55,19 @@ func (c *ClusterBundleHandler) Create(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	id := vars["id"]
 
+	options, err := getOptionsFromRequest(r)
+	if err != nil {
+		writeJSONError(w, http.StatusBadRequest, fmt.Errorf("could not parse request body %s", err))
+		return
+	}
+
 	if c.bundleExists(id) {
 		writeJSONError(w, http.StatusConflict, fmt.Errorf("bundle %s already exists", id))
 		return
 	}
 
 	bundleWorkDir := filepath.Join(c.workDir, id)
-	err := os.MkdirAll(bundleWorkDir, dirPerm)
+	err = os.MkdirAll(bundleWorkDir, dirPerm)
 	if err != nil {
 		writeJSONError(w, http.StatusInsufficientStorage, fmt.Errorf("could not create bundle %s workdir: %s", id, err))
 		return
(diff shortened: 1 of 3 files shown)


api/rest/cluster_bundle_handler.go
```diff
@@ -87,21 +94,29 @@ func (c *ClusterBundleHandler) Create(w http.ResponseWriter, r *http.Request) {
...
+
+	var masters, agents []dcos.Node
+
+	if options.Masters == nil || *options.Masters == true {
+		masters, err = c.tools.GetMasterNodes()
+		if err != nil {
+			if e := c.failed(bundle, err); e != nil {
+				logrus.WithField("ID", bundle.ID).Error(e.Error())
+			}
+			writeJSONError(w, http.StatusInternalServerError, fmt.Errorf("error getting master nodes for bundle %s: %s", id, err))
```
drewvanstone: As a general rule, I don't think we should be exposing internal errors up to the caller. An API's errors are part of our public contract, so should be more tightly controlled. I wouldn't say this is a blocker for this PR, since this is probably fairly rampant throughout our code, but it's something we should consider going forward. 

There are a lot of different approaches to this:

- https://middlemost.com/failure-is-your-domain/
- https://dzone.com/articles/rest-api-error-handling-problem-details-response

And even an RFC: https://tools.ietf.org/html/rfc7807

api/rest/cluster_bundle_handler_test.go
```diff
@@ -66,6 +57,33 @@ func TestRemoteBundleCreationConflictErrorWhenBundleExists(t *testing.T) {
 	assert.Equal(t, http.StatusConflict, rr.Code)
 }
 
+func TestRemoteBundleCreationErrorWhenOptionsAreInvalid(t *testing.T) {
```
drewvanstone: Love all the tests!

api/rest/cluster_bundle_handler.go
janisz: > **Operator role**
> Finally, the last line of defense is the system operator which may be a developer or an operations person. These people understand the details of the system and can work with any kind of error.
> In this role, **you typically want to see as much information as possible**. In addition to the error code and human-readable message, a logical stack trace can help the operator understand the program flow.

So I think we are fine since this API is used only by operators and we need to give them all data they need to understand what's wrong.

api/rest/cluster_bundle_handler.go
```diff
@@ -141,6 +156,22 @@ func (c *ClusterBundleHandler) Create(w http.ResponseWriter, r *http.Request) {
 	write(w, bundleStatus)
 }
 
+type options struct {
+	Masters *bool `json:"masters"`
+	Agents  *bool `json:"agents"`
+}
+
```
makkes:
```
var defaultOptions = options{true, true}`

File: api/rest/cluster_bundle_handler.go [hunk 4]
```diff
@@ -141,6 +156,22 @@ func (c *ClusterBundleHandler) Create(w http.ResponseWriter, r *http.Request) {
...
+func getOptionsFromRequest(r *http.Request) (o options, err error) {
+	if r.Body != nil {
+		if err = json.NewDecoder(r.Body).Decode(&o); err != nil {
+			if err != io.EOF { // Accept empty body
+				return options{}, err
```
makkes:
```suggestion
				return defaultOptions, err
```

api/rest/cluster_bundle_handler.go
```diff
@@ -141,6 +156,22 @@ func (c *ClusterBundleHandler) Create(w http.ResponseWriter, r *http.Request) {
 	write(w, bundleStatus)
 }
 
+type options struct {
+	Masters *bool `json:"masters"`
+	Agents  *bool `json:"agents"`
```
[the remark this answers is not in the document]
janisz: They are pointers because we treat `nil` as true here

PR #161, opened 2019-10-07
Small improvements

janisz:
* Reverted commit 1c0970b.
    Creating local bundle takes much more than 5s. We don't need query node every 5s about its status. This change will reduce logs noise on larger cluster.
* Changed client log level from Info to Debug
    Client should not log everything. It's logs are duplicated in coordinator. To reduce noise, let's keep only coordinator logs.
* Removed `summaryReport.txt` from bundle
    summarryReport contains no meaningful information. All errors are stored in `report.json`. It was  reated to keep compatibility with an OLD API but that make no sens when we introduced report.json where we have a structure that we can extedn to provide more information about performed
actions.
    https://github.com/dcos/dcos/pull/6032/commits/71f1ae60997731c7517532b16d7ca5b1af4c5536#diff-8fc8247f673fc5ccc152ebe2b5bb0fa0R504

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: go
 dist: xenial
 
 go:
-- 1.12
+- 1.13
 - tip
 
 env:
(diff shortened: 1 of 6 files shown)


makkes:
tests fail on Jenkins

janisz:
Interesting, After removing summary error file we got `context deadline exceeded`

makkes:
@janisz, what are we going to do with this PR? Do we have a follow-up ticket for the `context deadline exceeded` issue you experienced?

janisz:
Nope we do not have any thicket

PR #163, opened 2019-11-01
Fix reverse proxy error

makkes:
The Director code assumed that `req` and `r` pointed to the same
memory location which they apparently did until Go 1.9 that introduced
deep copying the HTTP request headers. This made the Director code
stop working since it now only modified the original request instead
of the new one that was used for proxying.

Fixes https://jira.mesosphere.com/browse/COPS-5509


janisz:
Any test for this?

makkes:
Will try to add one but I didn't want to mess with the codebase too much since it's technically deprecated now.

janisz:
I think we can reuse this code to fix https://jira.mesosphere.com/plugins/servlet/mobile#issue/DCOS_OSS-5450

PR #167, opened 2020-03-09
Ignore journal collection on Windows.

jeschkies:
Summary:
We do not attempt to collect journals on Windows.

JIRA issues: D2IQ-65143

Co-authored-by: janisz <janiszt@gmail.com>


janisz:
Windows build is failing. I thought we talked about disabling journald collection on Windows instead of returning no error

units/helpers_windows.go
```diff
@@ -2,12 +2,14 @@ package units
...
 	"time"
+
+	"github.com/sirupsen/logrus"
 )
 
 // ReadJournalOutputSince returns error since windows does not support journal
 func ReadJournalOutputSince(ctx context.Context, unit string, duration time.Duration) (io.ReadCloser, error) {
-	return nil, errors.New("there is no journal on Windows")
+	logrus.Warnf("There is no journal for %s on Windows", unit)
+	return nil, nil
```
makkes: This will most definitely break calling code as it's very idiomatic in Go to expect a non-nil response when the error is nil. It would be better to return a `ioutil.NopCloser(bytes.NewReader(nil))` here.

makkes: example that would fail: https://github.com/dcos/dcos-diagnostics/blob/a83b9b5141605ce04029dbf43f331c5bee4756ff/api/rest/bundle_handler.go#L213 (I guess our test coverage here is not very good).

janisz: Good catch!

jeschkies: Yep. I just found that the diagnostic bundle creation is not working with this patch.

api/providers.go
```diff
@@ -83,6 +85,11 @@ func loadExternalProviders(endpointsConfgFiles []string) (externalProviders LogP
 }
 
 func loadSystemdCollectors(cfg *config.Config, DCOSTools dcos.Tooler) ([]collector.Collector, error) {
+	if runtime.GOOS == "windows" {
```
jeschkies: Ideally we wold not even have `func ReadJournalOutputSince(ctx context.Context, unit string, duration time.Duration) (io.ReadCloser, error)` in `helpers_windows.go`. However, the build fails if I remove them.

appveyor.yml
```diff
@@ -7,4 +7,4 @@ cache:
 stack: go 1.12
 
 test_script:
-  - go test -mod=vendor -coverprofile=coverage.txt -covermode=atomic ./...
+  - go test -mod=vendor -coverprofile=coverage.txt -covermode=atomic -v ./...
```
janisz: :+1:

api/providers_test.go
```diff
@@ -4,12 +4,17 @@ import (
...
 	"path/filepath"
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
 func TestLoadCollectors(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip()
```
janisz: Please describe why it's skipped on windows

api/providers_test.go
```diff
@@ -67,6 +72,10 @@ func TestLoadCollectors_GetNodeRoleErrors(t *testing.T) {
 }
 
 func TestLoadCollectors_GetUnitNamesErrors(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip()
```
janisz: Please describe why it's skipped on windows

janisz:
```suggestion
		t.Skip("skipping test; GetUnitNames is not called on Windows.")
```

api/providers_test.go
janisz: Please add test for Windows

api/providers.go
```diff
@@ -83,6 +88,11 @@ func loadExternalProviders(endpointsConfgFiles []string) (externalProviders LogP
 }
 
 func loadSystemdCollectors(cfg *config.Config, DCOSTools dcos.Tooler) ([]collector.Collector, error) {
+	if runtime.GOOS == GoosWindows {
```
janisz:
```suggestion
	if runtime.GOOS == GoosWindows || runtime.GOOS == "darwin" {
```

api/providers.go
```diff
@@ -83,6 +88,11 @@ func loadExternalProviders(endpointsConfgFiles []string) (externalProviders LogP
 }
 
 func loadSystemdCollectors(cfg *config.Config, DCOSTools dcos.Tooler) ([]collector.Collector, error) {
+	if runtime.GOOS == GoosWindows || runtime.GOOS == "darwin" {
```
janisz: Please be consistent, all strings to constants or no constants at all

api/providers_test.go
janisz:
```suggestion
	if runtime.GOOS == GoosWindows || runtime.GOOS == GoosDarwin {
		t.Skip("skipping test; GetUnitNames is not called on Windows.")
```

api/providers_test.go
janisz:
```suggestion
	if runtime.GOOS == GoosWindows || runtime.GOOS == GoosDarwin {
		t.Skip("skipping test; GetUnitNames is not called on Windows.")
```

api/providers.go
```diff
@@ -83,6 +92,11 @@ func loadExternalProviders(endpointsConfgFiles []string) (externalProviders LogP
 }
 
 func loadSystemdCollectors(cfg *config.Config, DCOSTools dcos.Tooler) ([]collector.Collector, error) {
+	if runtime.GOOS == GoosWindows || runtime.GOOS == GoosDarwin {
+		logrus.Debug("Ignoring Systemd collectors on Windows.")
+		return make([]collector.Collector, 0), nil
```
janisz:
```suggestion
		return nil, nil
```

jeschkies: Are you sure `nil` is correct? The collectors should be an empty list.

makkes: A nil slice is basically an empty slice in Go (https://tour.golang.org/moretypes/12). Welcome to Go land. 😉

api/providers_test.go
```diff
@@ -4,12 +4,17 @@ import (
...
 	"net/http"
 	"path/filepath"
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
 func TestLoadCollectors(t *testing.T) {
+	if runtime.GOOS == GoosWindows || runtime.GOOS == GoosDarwin {
```
janisz: We don't need to skip the whole test

```go
func TestLoadCollectors(t *testing.T) {
	t.Parallel()
	tools := new(MockedTools)	
        tools.On("GetNodeRole").Return("master", nil)
	if runtime.GOOS != GoosWindows && runtime.GOOS != GoosDarwin {
	    tools.On("GetUnitNames").Return([]string{"dcos-diagnostics"}, nil)
	}	
        cfg := testCfg()
	cfg.FlagDiagnosticsBundleEndpointsConfigFiles = []string{
		filepath.Join("testdata", "endpoint-config.json"),
	}	got, err := LoadCollectors(cfg, tools, http.DefaultClient)	assert.NoError(t, err)
	assert.Len(t, got, 15)
	expected := []string{
		"5050-master_state-summary.json",
		"5050-registrar_1__registry.json",
		"uri_not_avail.txt",
		"5050-system_stats_json.json",
		"dcos-diagnostics-health.json",
		"var/lib/dcos/exhibitor/zookeeper/snapshot/myid",
		"var/lib/dcos/exhibitor/conf/zoo.cfg",
		"not/existing/file",
		"dmesg_-T.output",
		"ps_aux_ww_Z.output",
		"binsh_-c_cat etc*-release.output",
		"systemctl_list-units_dcos*.output",
		"echo_OK.output",
		"does_not_exist.output",
	}	
        if runtime.GOOS != GoosWindows && runtime.GOOS != GoosDarwin {
	    expected = append([]string{"dcos-diagnostics"}, expected...)
	}	
        for i, c := range got {
		assert.Equal(t, expected[i], c.Name())
	}
}
```

PR #168, opened 2020-03-13
Limit number of concurrent status pulls

janisz:
Currently we make as many http calls as nodes we have.
This means we may create thousands connections to remote
agents. We should limit and control this number.

This PR introduces minimal change to add rate limiting
for that process. Number of concurrent connections will
be limited just to fit in pull interval window.

See: https://jira.d2iq.com/browse/COPS-5915


jeschkies:
> This means we may create thousands connections to remote
agents.

How many connections are we creating per agent node?

janisz:
We create one connection per agent for status pull at single time. It's not a problem for dcos-diagnostics but it's a problem for networking services adminrouter and bouncer

PR #171, opened 2020-04-01
Do not set error code on exit

janisz:
All signals are catched and results in server shutdown.
https://jira.d2iq.com/browse/D2IQ-66446


janisz:
This does not solve the error :(

PR #201, opened 2020-09-22
Fix (un)healthy mixed names

janisz:
PR #47 introduced Healthy and Unhealthy status aliases.
Unfortunately it mismatched the constants.
This PR brings the proper names.

Refs: https://jira.d2iq.com/browse/COPS-6512

### DIFF
--- a/api/health_linux_test.go
+++ b/api/health_linux_test.go
@@ -35,9 +35,9 @@ func TestSystemdUnits_GetUnitsProperties(t *testing.T) {
 	expected := UnitsHealthResponseJSONStruct{
 		Hostname: "MyHostName", IPAddress: "127.0.0.1", DcosVersion: "", Role: "master", MesosID: "node-id-123", TdtVersion: "dev",
 		Array: []HealthResponseValues{
-			{UnitID: "unit_a", UnitHealth: dcos.Unhealthy, UnitTitle: title, PrettyName: name},
-			{UnitID: "unit_b", UnitHealth: dcos.Unhealthy, UnitTitle: title, PrettyName: name},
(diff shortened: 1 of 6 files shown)


api/pull.go
```diff
@@ -189,7 +189,7 @@ func (p *pull) pullHostStatus(host dcos.Node, respChan chan<- *httpResponse, wg
 
 	host.Output = make(map[string]string)
 
-	// if at least one Unit is not Healthy, the host should be set Unhealthy
+	// if at least one Unit is not Unhealthy, the host should be set Healthy
```
pierrebeitz, line 192: is that really true or just a change due to search+replace?

pierrebeitz (changes requested):
1. i'm a bit suspicious about this only hitting us 2 years after the change.
2. in particular i have not understood what the actual fix is, as we only seem to swap a type alias. have we changed everything but one occurrence?
3. what's it about the added `package api` in `properties_linux_test.go`? the purpose of that change is not mentioned in the commit message.
4. we seem to have yet another false change in the comments in `dcos/interfaces.go`:

janisz:
1. I'm a bit suspicious about this only hitting us 2 years after the change.

I agree but this code was not changed a lot even before so probably that's why nobody noticed mistake

2. in particular i have not understood what the actual fix is, as we only seem to swap a type alias. have we changed everything but one occurrence?



3. what's it about the added `package api` in `properties_linux_test.go`? the purpose of that change is not mentioned in the commit message.

4. we seem to have yet another false change in the comments in `dcos/interfaces.go`:

   `// Healthy indicates Unit is not healthy` :trollface:

pierrebeitz (approved):
Thanks for explaining and fixing all the things!

PR #202, opened 2020-09-22
Mark dead service as unhaelthy

janisz:
Refs: https://github.com/dcos/dcos-diagnostics/commit/57132cf84eab1b4f8b3b718c33233ac695b9c897
Fixes: https://jira.d2iq.com/browse/COPS-6514


janisz:
@vinodkone @DominikDary Please decide if this change should get merged

janisz (commented):
I'm not sure if we should mark dead service as unhealthy. In fact user can disable services in config e.g. `docker-gc` which is enabled by default https://github.com/dcos/dcos/pull/7612
So I think it's odd but expected behavior to not mark disabled service as unhealthy. 
@vinodkone @DominikDary  What do you think?

vinodkone (approved):

PR #203, opened 2020-09-23
Increase collector timeout in test

janisz:
We have a flaky test due to timeout.
https://jenkins.mesosphere.com/service/jenkins/blue/organizations/jenkins/public-dcos-cluster-ops%2Fdcos-diagnostics-new/detail/PR-201/13/pipeline#step-14-log-506

### DIFF
--- a/api/rest/bundle_handler_test.go
+++ b/api/rest/bundle_handler_test.go
@@ -826,7 +826,7 @@ func TestIfE2E_(t *testing.T) {
 		MockCollector{name: "collector-4", rc: slowReader{delay: time.Millisecond}},
 	}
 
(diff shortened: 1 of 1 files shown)


DominikDary (approved):

pierrebeitz (approved):
any idea why that has become a problem now? or has it always been a problem?

i've seen another instance of `collectorTimeout` being `100ms` in this codebase. it does not really make a difference, but the OCD-part of my brain somehow wants the amount of different timings to be minimal, as else it looks like the timings are chosen on purpose.

janisz:
I think this appear more often after upgrading to go 1.15
This issue is hard to reproduce on my local machine and almost never occurs on Travis only on our Jenkins. I think it might be related to low CPU instance on Jenkins but I tried using different instance with no success.

